PR #2, opened 2022-12-23
Escapes TS union symbol in PanelGroup type

dylanklohr:
## Description

Fixes a display issue for `PanelGroup` prop types where the TS union of string literals `"horizontal"` and `"vertical"` was erroneously considered a markdown column delimiter.

**Before**

ScreenShot2022-12-22 at 19 23 01@2x

**After**

ScreenShot2022-12-22 at 19 23 31@2x


bvaughn (approved):
Thx

PR #11, opened 2022-12-24
Basic touch support

andskr:
Translate touch event to movement X and Y coordinates, so then we can reuse them to calculate movement in the same way as we do for MouseEvent

We should consider disabling the user selection while users resize panels (you can see that I can accidentally select some test inside panned while dragging the handle on mobile screen recording).

https://user-images.githubusercontent.com/61060964/209440016-b7875071-f9ad-4d58-a214-48ab8d430162.mp4

https://user-images.githubusercontent.com/61060964/209440035-005266b1-8669-4390-a3de-46c261555c9e.mp4

### DIFF
--- a/packages/react-resizable-panels/src/PanelGroup.tsx
+++ b/packages/react-resizable-panels/src/PanelGroup.tsx
@@ -11,7 +11,7 @@ import {
 import useUniqueId from "./hooks/useUniqueId";
 
 import { PanelContext, PanelGroupContext } from "./PanelContexts";
-import { Direction, PanelData } from "./types";
+import { Direction, PanelData, ResizeEvent } from "./types";
 import { loadPanelLayout, savePanelGroupLayout } from "./utils/serialization";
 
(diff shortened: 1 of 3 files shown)


bvaughn:
This looks really nice! 👏🏼  Thank you!

> We should consider disabling the user selection while users resize panels

Agreed! Maybe we could do this at the same place as we set the document-level resize cursor (in the `PanelResizeHandler` effect)?

bvaughn:
Just tried this on my iPhone and (aside from the drag handle touch area being too small– which is just my demo's fault) it seems to work nicely!

andskr:
> Just tried this on my iPhone and (aside from the drag handle touch area being too small– which is just my demo's fault) it seems to work nicely!

Yeah, I also saw that, and just increased the handle size to test it

bvaughn:
Nice nice! I'm going ot make a few subjective nits to naming things and then I'll merge this!

PR #34, opened 2022-12-27
Fix button text

pawelblaszczyk5:
Both buttons text was dependent on one variable, instead of separate

Have a nice day!


bvaughn:
Oh, thank you for this! I actually ended up fixing this as part of #33 but thank you!

PR #37, opened 2022-12-27
fix(website): add viewport to make it scalable based on device width

jyash97:
Testing on mobile needed zoom in, thats why added a quick fix


bvaughn (approved):
Thx~

PR #38, opened 2022-12-27
Add [data-resize-handle-active] attribute to resize panels

### LINKED ISSUE #36: Active resize handle styling
If you use the website in Mobile and try resizing panels you will notice that the active styles are not triggered, on debugging I found that on the web it focuses on the handle but in Mobile, the focus is only added if we click the handle but if we start moving it doesn't focus on the handle. 

Maybe we can add a data attribute called `data-panel-active=true` whenever the handle is active and is being dragged. As of now we need to add focus styles for the active handle.

I am open to helping on this issue :D

Reproducible steps:
https://user-images.githubusercontent.com/22376783/209684416-e780f949-4a5d-4008-8c66-09fd61a2c109.mov

bvaughn:
Remove `PanelContext` from public API

Active handle sizing can now be done via `[data-resize-handle-active]` with pure CSS, e.g.:
```css
.ResizeHandleOuter {
  /* Base style */
}
.ResizeHandleOuter[data-resize-handle-active] {
  /* Active style */
}
```

Resolves #36

### DIFF
--- a/packages/react-resizable-panels-website/src/components/ResizeHandle.module.css
+++ b/packages/react-resizable-panels-website/src/components/ResizeHandle.module.css
@@ -1,23 +1,21 @@
-.ResizeHandle {
+.ResizeHandleOuter {
   flex: 0 0 1em;
   position: relative;
   outline: none;
+
+  --background-color: transparent;
+}
+.ResizeHandleOuter[data-resize-handle-active] {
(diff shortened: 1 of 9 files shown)


jyash97 (approved):
Works perfectly ✌️

bvaughn:
Fix publshed as 0.0.16

PR #44, opened 2022-12-29
Don't cancel resize operations when exiting the window

### LINKED ISSUE #42: Ability to keep dragging with mouse outside the window
If you compare https://codesandbox.io/s/react-resizable-panels-zf7hwd and https://react-grid-layout.github.io/react-grid-layout/ you'll notice that `react-grid-layout` lets you keep dragging even if your mouse leaves the window, but with `react-resizable-panels` the drag event seems to stop as soon as the mouse leaves the window.

This is probably not a dealbreaker, but it would arguably be nicer UX if the drag event didn't just stop if you accidentally dragged too much.

Sorry, I'm currently on my phone so I can't record a video to make it crystal clear what I'm talking about but I hope the description suffices!

bvaughn:
Only cancel when a `"mouseup"` (or `"touchend"`) event is fired.

Resolves #42

### DIFF
--- a/packages/react-resizable-panels/src/PanelResizeHandle.ts
+++ b/packages/react-resizable-panels/src/PanelResizeHandle.ts
@@ -92,19 +92,19 @@ export default function PanelResizeHandle({
     };
 
     document.body.addEventListener("contextmenu", stopDraggingAndBlur);
-    document.body.addEventListener("mouseleave", stopDraggingAndBlur);
     document.body.addEventListener("mousemove", onMove);
     document.body.addEventListener("touchmove", onMove);
-    document.body.addEventListener("mouseup", stopDraggingAndBlur);
+    window.addEventListener("mouseup", stopDraggingAndBlur);
+    window.addEventListener("touchend", stopDraggingAndBlur);
 
     return () => {
       document.body.style.cursor = "";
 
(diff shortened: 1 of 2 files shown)


packages/react-resizable-panels/src/PanelResizeHandle.ts
```diff
@@ -92,19 +92,19 @@ export default function PanelResizeHandle({
...
     return () => {
       document.body.style.cursor = "";
 
       document.body.removeEventListener("contextmenu", stopDraggingAndBlur);
-      document.body.removeEventListener("mouseleave", stopDraggingAndBlur);
       document.body.removeEventListener("mousemove", onMove);
       document.body.removeEventListener("touchmove", onMove);
-      document.body.removeEventListener("mouseup", stopDraggingAndBlur);
+      window.removeEventListener("mouseup", stopDraggingAndBlur);
+      window.removeEventListener("touchend", stopDraggingAndBlur);
```
NMinhNguyen, line 107: Not for right now, but I wonder if you'll need to eventually derive (or accept) the correct `document` and `window` to support rendering into iframes / child windows opened via `window.open()`

bvaughn: Maybe! I guess it depends on where this thing gets used. This works for now 😁

PR #50, opened 2022-12-31
Improve global resize cursor [help wanted]

bvaughn:
When the mouse nears the browser window boundary, the cursor is updated (by the OS?) and does not get reset when it 
re-enters. This commit attempts to track and recover from this state. For the most part it seems to work okay:

Video showing cursor changing near the window boundary and resetting when re-entering the window

Occasionally though the cursor won't be reset correctly:

Edge case demonstrating cursor not resetting after re-entering the window

I'm on the fence about whether this change is worth making at all– especially if it has edge cases. Suggestions welcome!

### DIFF
--- a/packages/react-resizable-panels/src/utils/cursor.ts
+++ b/packages/react-resizable-panels/src/utils/cursor.ts
@@ -6,8 +6,11 @@ type CursorState =
   | "vertical-max"
   | "vertical-min";
 
+const WINDOW_EDGE_THRESHOLD = 25;
+
 let currentState: CursorState | null = null;
 let element: HTMLStyleElement | null = null;
+let previousState: CursorState | null = null;
 
 export function getCursorStyle(state: CursorState): string {
   switch (state) {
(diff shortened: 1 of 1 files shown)


shshaw:
One trick I've had to do for resizing/moving is using a full window overlay. This helps prevent other elements from blocking the mouse/touch move events.

1. On mousedown, create/reveal a `position: fixed; inset: 0; z-index: 9999` overlay element. 
2. Attach the move and end events to this overlay element
3. Destroy/hide the overlay element on end/exit.

If you apply your cursor style to this overlay element, that'll ensure the cursor stays consistent until the overlay is destroyed.

shshaw:
Another option is `setPointerCapture` which "locks" the pointer events to that element.

https://developer.mozilla.org/en-US/docs/Web/API/Element/setPointerCapture

```
  function onDown(e) {
    e.preventDefault();
    // Capture pointer events to prevent moving outside from the element
    // NOTE: There may be an issue in Safari with this. Needs deeper testing.
    // https://twitter.com/WestbrookJ/status/1515677167262351362
    // https://codepen.io/joyzhong/pen/ZEpvdEx
    el.current.setPointerCapture(e.pointerId);
```

bvaughn:
Interesting approaches! Would either of those fix the OS cursor issue referenced in the description, I wonder? For example, I think the overlay approach would be similar to setting the style on the HTMLBodyElement.

shshaw:
Quick testing with `setPointerCapture` persists the cursor outside of the window. That should be easy to drop in to your event listeners, and then you'd set the cursor on the element itself to get the proper cursor.

bvaughn:
Very cool 🥳

I'll take a look at this sometime in the evening.

shshaw:
Was taking a stab at doing a simple `stopPointerCapture()` implementation but wasn't able to get far before running into issues due to my unfamiliarity with things. Ran out of time to try it, but I did want to pass on:

1. You have to use `pointerdown` and `pointerup`/`pointercancel` as the main events listeners in order for `setPointerCapture` to be usable. That does change browser support a bit, but basically everything over 4 years old (Safari 13+) supports it. https://caniuse.com/mdn-api_element_setpointercapture

2. You may also be able to drop all the `document` and `window` listeners to instead use `pointermove` events on the element itself, further simplifying things.

bvaughn:
Thanks for sharing!

bvaughn:
> Quick testing with `setPointerCapture` persists the cursor outside of the window.

Just took a look at this in [code sandbox](https://codepen.io/shshaw/pen/zYLKGje) and it doesn't seem to be the case for me.


https://user-images.githubusercontent.com/29597/211154677-6da1c0cd-448c-4a5f-85cd-209078a6dcd4.mp4

shshaw:
> Just took a look at this in [code sandbox](https://codepen.io/shshaw/pen/zYLKGje) and it doesn't seem to be the case for me.

Odd. It works for me in Chrome & Firefox on macOS. Safari seems to stop outside of the frame, but if you keep holding down the event persists when you get back in the frame.

https://user-images.githubusercontent.com/777155/212076833-9a5fb590-9036-4e1e-8540-fff57c44575a.mp4

Are you trying it in that CodePen link, or on CodeSandbox? If it's in an iframe, the `sandbox="allow-pointer-lock"` attribute has to be there. I'm not sure if CodeSandbox does that.

Screenshot 2023-01-12 at 6 55 50 AM

PR #57, opened 2023-01-03
fix: update server check to use document

Mokshit06:
Server runtimes such as Deno have `window` global, but they should still use `useEffect` and not `useLayoutEffect`. `document` is a better way to check if it is a server runtime or browser


bvaughn:
Maybe it's best to just mirror [React's own check](https://github.com/facebook/react/blob/main/packages/shared/ExecutionEnvironment.js) for this:
```js
!!(
  typeof window !== 'undefined' &&
  typeof window.document !== 'undefined' &&
  typeof window.document.createElement !== 'undefined'
)
```

bvaughn:
Released as 0.0.29

Thanks!

PR #61, opened 2023-01-05
Adds copy button

### LINKED ISSUE #60: Copy button for code snippits
Ideally, the code snippets should have copy buttons. You can copy the code now by hand, but you end up getting the line numbers as well, which is annoying to try to take out.

Proposed fix in PR #61

mackenly:
Adds a copy button to the code snippets on the website. 

Closes #60 

Known issue: Because of how the overflow blocks work and the Code component builds the lines, I've been unable to find a good solution for integrating the button without refactoring the whole Code component. The copy button is hidden on the overflow panels as a temporary fix.


bvaughn:
Hey thanks for the PR. I appreciate it!

I think I'd rather avoid adding a _copy_ button though. Feels a bit overkill for small code snippets like this.

The line thing mentioned in #60 is understandably annoying, but we could fix that by setting `user-select: none` on the `LineNumber` class.

PR #76, opened 2023-01-10
Panel and PanelGroup callbacks always called after mount

### LINKED ISSUE #75: Panel onResize and onCollapse props should be called after mount
Currently these callbacks are only triggered after a user-initiated resize. Combining this with the auto-save behavior leads to an ambiguous mount scenario: is the panel collapsed or not? what's the initial size?

bvaughn:
- [x] `PanelGroup` should call `onLayout` exactly once, after mounting with calculated sizes.
- [x] `Panel` should call `onCollapse` and `onResize` exactly once, after mounting with calculated sizes.
- [x] e2e tests added for the above cases.

Resolves #75


bvaughn:
Released in 0.0.32

PR #80, opened 2023-01-12
onCollapse props is not called after mount

jeetiss:
- `onCollapse` doesn't fire on mount with `v0.0.32`

- `callPanelCallbacks` is fired with an empty array as a `prevProps`

https://github.com/bvaughn/react-resizable-panels/blob/1696b4f2eabaf20b1ee71f4128acda2000eb9ef9/packages/react-resizable-panels/src/PanelGroup.ts#L139

- `prevSize` equal to `undefined` here

https://github.com/bvaughn/react-resizable-panels/blob/1696b4f2eabaf20b1ee71f4128acda2000eb9ef9/packages/react-resizable-panels/src/utils/group.ts#L101

- `onCallapsed` is not called because `prevSize` is undefined

https://github.com/bvaughn/react-resizable-panels/blob/1696b4f2eabaf20b1ee71f4128acda2000eb9ef9/packages/react-resizable-panels/src/utils/group.ts#L111-L114

---

I tried to run tests but it seems them doesn't work so i didn't write any test for this case


bvaughn:
> `onCollapse` doesn't fire on mount with v0.0.32

This function only fires if the panel is _collapsed_ after mounting. (It doesn't fire for panels that aren't collapsed.) That's the behavior that made sense to me, but it seems like you're suggesting that's not the expected behavior for you?

bvaughn:
It looks like b5d3ec1 caused a regression in a few of the tests and I didn't notice. (Whoops!)

jeetiss:
> This function only fires if the panel is collapsed after mounting

I just missed this information for some reason, though `onCollapse` would fire as `onResize` on every mount. thanks for update

bvaughn:
Nah, it's okay. I think it's reasonable for it to fire with an initial "false" value too.

bvaughn:
#81

bvaughn:
v0.0.33

PR #87, opened 2023-01-17
Fix cursor when using keyboard

### LINKED ISSUE #86: Cursor not changing back to default cursor after using keyboard to control panel resize handle
The cursor does not change back to the default cursor after using the keyboard to reposition the resize handle. It only changes back after having used the cursor to reposition the resize handle. Until then, it stays on `col-resize` and only changes back to `default` after using the cursor.

Video of this happening:

papahigh:
Closes #86
--
Reset cursor style on blur for PanelResizeHandle

### DIFF
--- a/packages/react-resizable-panels-website/tests/CursorStyle.spec.ts
+++ b/packages/react-resizable-panels-website/tests/CursorStyle.spec.ts
@@ -1,0 +1,35 @@
+import { expect, test } from "@playwright/test";
+import { createElement } from "react";
+import { Panel, PanelGroup, PanelResizeHandle } from "react-resizable-panels";
+
+import { goToUrl } from "./utils/url";
+
+test.describe("cursor style", () => {
+  function getCursorStyle(element: Element) {
+    return window.getComputedStyle(element).getPropertyValue("cursor");
+  }
+
+  test("should reset cursor style on resize handle blur", async ({ page }) => {
+    await goToUrl(
+      page,
+      createElement(
+        PanelGroup,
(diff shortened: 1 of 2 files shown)


bvaughn (commented):
Thank you for posting this! 🙇🏼 Also _love_ to see the added test!

I wonder if a slightly better approach to this might be to avoid setting a global cursor style _entirely_ for keyboard related interactions?

I think we could accomplish this in the `resizeHandler` function (in `PanelGroup`) with a check like...

```tsx
if (isMouseEvent(event) || isTouchEvent(event)) {
  if (prevSizes === nextSizes) {
    // If the pointer has moved too far to resize the panel any further,
    // update the cursor style for a visual clue.
    // This mimics VS Code behavior.

    if (isHorizontal) {
      setGlobalCursorStyle(
        movement < 0 ? "horizontal-min" : "horizontal-max"
      );
    } else {
      setGlobalCursorStyle(
        movement < 0 ? "vertical-min" : "vertical-max"
      );
    }
  } else {
    // Reset the cursor style to the the normal resize cursor.
    setGlobalCursorStyle(isHorizontal ? "horizontal" : "vertical");
  }
}

if (prevSizes !== nextSizes) {
  // If resize change handlers have been declared, this is the time to call them.
  callPanelCallbacks(panelsArray, prevSizes, nextSizes);

  setSizes(nextSizes);
}
```

bvaughn:
I'm going to pick up where you've started and make the above tweak: #88

Issue #72, opened 2023-01-10
Module parse error


nouman91:
I am trying to using the library in an existing project bootstrapped with craco and create-react-app. Following is the error I get at the compile time. 

```
./node_modules/react-resizable-panels/dist/react-resizable-panels.module.js 141:23
Module parse failed: Unexpected token (141:23)
File was processed with these loaders:
 * ./node_modules/babel-loader/lib/index.js
You may need an additional loader to handle the result of these loaders.
|   if (state) {
|     const key = $6ff1a22b27cc039d$var$getSerializationKey(panels);
>     return state[key] ?? null;
|   }
|
```

Package.json details:
`
"@craco/craco": "^6.0.0",
"react": "^17.0.2",
 "react-dom": "^17.0.2",
`

Please let me know if you need more details.

bvaughn:
Duplicate of #67 

Should find a solution here:
https://github.com/facebook/create-react-app/issues/9468#issuecomment-982471433

Good luck! 😄

bvaughn:
FYI, I removed the nullish coalescing operator in v0.0.31 because the trivial benefit it provided didn't seem worth causing headaches for create-react-app users.

nouman91:
Thank you for your prompt response. After a day of fiddling around I could not get it to work here is error which I am getting. 

`Failed to compile.

./node_modules/react-resizable-panels/dist/react-resizable-panels.module.js 316:38
Module parse failed: Unexpected token (316:38)
File was processed with these loaders:
 * ./node_modules/babel-loader/lib/index.js
You may need an additional loader to handle the result of these loaders.
|   const handles = $f30c804b5fe6cc1a$export$ae14931f0a0256a3(groupId);
|   const index = handles.indexOf(handle);
>   const idBefore = panelsArray[index]?.id || null;
|   const idAfter = panelsArray[index + 1]?.id || null;
|   return [idBefore, idAfter];`

bvaughn:
Sorry @nouman91. You'll probably need to look at Stack Overflow or somewhere else for help with the CRA Webpack configuration. That's not something I have the time or knowledge to support.

nouman91:
Yeah I understand! Thank you!

nouman91:
I was able to solve it, if someone is else is stuck into same issue with craco. Following is the package you need
https://www.npmjs.com/package/craco-babel-loader

bvaughn:
[...]

PR #92, opened 2023-01-18
set browerslist to 'Chrome 79'

reinhard:
By setting the browserlist to 'Chrome 79', Parcel produces Javascript that can be parsed with the Babel version available in CRA 4.0.3. This addresses problems with optional chaining, which is available starting with Chrome 80 (see https://caniuse.com/mdn-javascript_operators_optional_chaining)

See also https://github.com/bvaughn/react-resizable-panels/issues/72

### DIFF
--- a/packages/react-resizable-panels/package.json
+++ b/packages/react-resizable-panels/package.json
@@ -23,5 +23,8 @@
   "peerDependencies": {
     "react": "^16.14.0 || ^17.0.0 || ^18.0.0",
     "react-dom": "^16.14.0 || ^17.0.0 || ^18.0.0"
-  }
+  },
+  "browserslist": [
+    "Chrome 79"
+  ]
 }


bvaughn (approved):
Ah, thanks! Spot checking this.

Before this change, this line:
https://github.com/bvaughn/react-resizable-panels/blob/f93ca94007258bd155218b77791c777c35944a89/packages/react-resizable-panels/src/utils/group.ts#L243

Ended up in the bundle as:
```js
const idBefore = panelsArray[index]?.id || null;
```

With this change it becomes:
```js
const idBefore =
  ((_panelsArray_index = panelsArray[index]) === null ||
  _panelsArray_index === void 0
    ? void 0
    : _panelsArray_index.id) || null;
```

And if we revert 8677dab742347e533c1b499d3256e55c32950277 (on top of this) it becomes:
```js
const idBefore =
  (_panelsArray_index_id =
    (_panelsArray_index = panelsArray[index]) === null ||
    _panelsArray_index === void 0
      ? void 0
      : _panelsArray_index.id) !== null && _panelsArray_index_id !== void 0
    ? _panelsArray_index_id
    : null;
```

bvaughn:
Published as 0.0.35

https://github.com/bvaughn/react-resizable-panels/releases/tag/v0.0.35

reinhard:
Thanks for the ultra-fast merge (and the great component).

PR #95, opened 2023-01-25
Removed useId import that is not compatible with React < 18

thomas-milburn:
When using React versions under 18, useId is not exported by React which causes this error

`export 'useId' (imported as '$jhddX$useId') was not found in 'react'`

This pull removes the dependency on useId

### DIFF
--- a/packages/react-resizable-panels/src/hooks/useUniqueId.ts
+++ b/packages/react-resizable-panels/src/hooks/useUniqueId.ts
@@ -1,13 +1,11 @@
-import { useId, useRef } from "react";
+import { useRef } from "react";
 
 let counter = 0;
 
 export default function useUniqueId(
   idFromParams: string | null = null
 ): string {
-  const idFromUseId = typeof useId === "function" ? useId() : null;
-
-  const idRef = useRef<string | null>(idFromParams || idFromUseId || null);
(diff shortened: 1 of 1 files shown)


bvaughn:
I think this "error" you're reporting has something to do with your build tooling.

 I've tested this package with React 18, 17, and 16 on Code Sandbox, and it works without any errors. There is an is-defined check in the code specifically to handle older versions of React:
```js
const idFromUseId = typeof useId === "function" ? useId() : null;
```

Removing `useId`, as this PR does, would break server rendering for users that don't explicitly provide `id` props.

thomas-milburn:
Thanks for the quick response and the explanation. I'll have a look at my webpack configuration as I'm just using the default one from create react app.

I also have a follow-up question, if UUID's were used instead of a counter would this fix the server rendering issues? Thanks again for all your help!

bvaughn:
Gotcha. Unfortunately I don't have the bandwidth to help debug what's going on in CRA. 😢 

> if UUID's were used instead of a counter would this fix the server rendering issues? Thanks again for all your help!

Not if you're referring to auto-generated ones, because they wouldn't be guaranteed to match between server and client. That's the main function of [`useId`](https://beta.reactjs.org/reference/react/useId).

thomas-milburn:
No worries, thanks again for the help :)

PR #107, opened 2023-02-21
Issue/103

ethanyou725:
Add affect direction to Panel imperative api collapse/expand
```
type AffectDirection = 'before' | 'after';
collapse: (direction?: AffectDirection) => void;
expand: (direction?: AffectDirection) => void;
```
It means a Panel when be collapsed by imperative api, it will affect the sibling panel's size before or after it, default is after.


bvaughn:
Thank you for this proposal 🙏🏼  After thinking on it a little bit, I've decided that I don't think I want to add this functionality to the library. I understand the use case, but I think the "before" and "after" API is a bit awkward and I think I'd prefer to just stick with the current constraints.

I hope you'll understand 🙇🏼  and you are definitely welcome to fork this module if you'd prefer to build on top of this proposal for your own application.

PR #115, opened 2023-03-30
Dont interfere with outside handlers

floscr:
When focusing on the resizehandler and resizing with the keyboard, outside event keyhandlers would be executed as well.

### DIFF
--- a/packages/react-resizable-panels/src/hooks/useWindowSplitterBehavior.ts
+++ b/packages/react-resizable-panels/src/hooks/useWindowSplitterBehavior.ts
@@ -82,6 +82,7 @@ export function useWindowSplitterPanelGroupBehavior({
       handle.setAttribute("aria-valuenow", "" + Math.round(parseInt(flexGrow)));
 
       const onKeyDown = (event: KeyboardEvent) => {
+        event.stopPropagation();
         switch (event.key) {
           case "Enter": {
             const index = panelsArray.findIndex(
--- a/packages/react-resizable-panels/src/hooks/useWindowSplitterBehavior.ts
+++ b/packages/react-resizable-panels/src/hooks/useWindowSplitterBehavior.ts
@@ -82,9 +82,10 @@ export function useWindowSplitterPanelGroupBehavior({
       handle.setAttribute("aria-valuenow", "" + Math.round(parseInt(flexGrow)));
 
       const onKeyDown = (event: KeyboardEvent) => {
-        event.preventDefault();
         switch (event.key) {
           case "Enter": {
+            event.preventDefault();
+
             const index = panelsArray.findIndex(
               (panel) => panel.id === idBefore
             );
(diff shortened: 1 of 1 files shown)


packages/react-resizable-panels/src/hooks/useWindowSplitterBehavior.ts
```diff
@@ -82,6 +82,7 @@ export function useWindowSplitterPanelGroupBehavior({
       handle.setAttribute("aria-valuenow", "" + Math.round(parseInt(flexGrow)));
 
       const onKeyDown = (event: KeyboardEvent) => {
+        event.stopPropagation();
```
bvaughn, line 85: I don't think this is the right fix.

_Maybe_ the "Enter" case below should call `event.preventDefault()` to signal that it's handled the event, but I don't think it the component should swallow all keyboard events. Can you explain why you think that change is needed?

floscr: My issue is that I have an key up listener outside of the component like here:
https://codesandbox.io/s/react-resizable-panels-forked-vywou9

When resizing via the keyboard, the outside listener get's triggered as well.

`event.preventDefault()` would not resolve the issue.

Another option would be to be able to pass in a custom onKeyUp handler from the outside and handle it from the consumer side?

bvaughn: > `event.preventDefault()` would not resolve the issue.

Your key handler would need to respect `event.defaultPrevented`

floscr: Interesting, I've never encountered this approach.
It works, but it's hard to introduce this change in a giant code base, where this wasn't previously considered (which is of course not your responsibility, just trying to provide context :smile_cat: ).

Right now this is blocking me from using this library, as I can't even disable the focus ability or change the handlers.

bvaughn: I think [`defaultPrevented`](https://developer.mozilla.org/en-US/docs/Web/API/Event/defaultPrevented) would be a reasonable strategy here to signal that the event has already been "handled". Alternately you could block this event _in your own code_ by wrapping the resize handle, e.g.

```jsx
<div onKeyDown={trapEvent}>
  <PanelResizeHandle {...props} />
</div>
```

floscr: Ah that's also a solution, thank you!
I'm investigating how to include `defaultPrevented` and think it's a good way to go, I've changed the line in the PR now!

bvaughn (approved):
I think we shouldn't claim events as handled (`event.preventDefault`) unless we actually handle them.

bvaughn (approved):

bvaughn:
Oof. Looks like that commit broke some of the e2e tests. Going to re-open until I figure out what happened.

bvaughn:
Published in 0.0.38

Issue #48, opened 2022-12-30
Flickering slider saveId bug


abrman:
The flickering: https://codesandbox.io/s/react-resizable-panels-forked-r68rvr?file=/src/App.js

I've just spent over an hour trying to pinpoint why my sliders were flickering.
Long story short, I had the slider state saved in the local storage with the following content:
```
{"1:10":[20],"1:10,1:10":[10,30],"1:10,2:10":[16.461173681090695,23.538826318909305],"1:10,2:10,3:10":[18.993775933609953,31.006224066390047,10],"1:10,3:10":[10,30],"1:100,2:10,3:10":[10,40,10],"1:5,2:10,3:10":[10,40,10],"10,10":[16.611570247933884,23.388429752066116],"10,10,10":[0,66.66666666666667,33.333333333333336],"0,10,10":[0,61.74632352941176,38.25367647058824]}"
```

Deleting the local storage entry fixed my problem.

I don't know how it became broken - I worked off of the codesandbox example so occasionally I had three panels, maybe 4 at most. I'm not sure exactly what the JSON means, but I'm dropping the content of it in here hoping someone will understand what happened, and will help avoid the bug from occurring again. (Not sure whether it's exclusive to development or can also occur in production)

bvaughn:
If you figure out the cause, or a repro, let me know.

abrman:
From: https://github.com/bvaughn/react-resizable-panels/issues/47#issuecomment-1368100763
> You should be able to accomplish this using `minSize`/`maxSize` and a min/max external `height`/`width` though!

I wasn't sure what you meant with min/max external `height`/`width` and ran into the bug gain.
I suppose this is how you reproduce it: https://codesandbox.io/s/react-resizable-panels-forked-d8eqci?file=/src/App.js

Would you be willing to explain more of how you'd go about achieving min/max limits?

abrman:
I suppose I triggered the bug previously with setting `max-width` on one of the panels previously. It behaves similarly to the beforementioned sandbox link.

bvaughn:
Ah, right. I don't think I'd consider that a bug in this library (or something I'd try to fix). It's just kind of an invalid configuration.

bvaughn:
> Would you be willing to explain more of how you'd go about achieving min/max limits?

If you want to set min/max sizes _in pixels_ and you control the overall width and height of the panel group, then you can do the conversion to whatever the min/max percentage would be.

bvaughn:
[...]

PR #146, opened 2023-05-14
Add getElement() functions to Panel and PanelGroup

bartlangelaan:
This makes it much easier to do stuff like in https://github.com/bvaughn/react-resizable-panels/issues/48#issuecomment-1368106118, because you don't need a querySelector anymore but can use the ref to target the group or panel directly.

Also, this is the minimal implementation needed to get this working (I think), feel free to adjust & fix anything I missed.

### DIFF
--- a/packages/react-resizable-panels/src/Panel.ts
+++ b/packages/react-resizable-panels/src/Panel.ts
@@ -37,6 +37,7 @@ export type PanelProps = {
 };
 
 export type ImperativePanelHandle = {
+  getElement: () => HTMLElement | null;
   collapse: () => void;
   expand: () => void;
   getCollapsed(): boolean;
@@ -154,9 +155,12 @@ function PanelWithForwardedRef({
     };
   }, [registerPanel, unregisterPanel]);
 
+  const elementRef = useRef<HTMLElement>(null);
+
   useImperativeHandle(
     forwardedRef,
     () => ({
+      getElement: () => elementRef.current,
       collapse: () => collapsePanel(panelId),
       expand: () => expandPanel(panelId),
       getCollapsed() {
@@ -182,6 +186,7 @@ function PanelWithForwardedRef({
       ...style,
       ...styleFromProps,
     },
+    ref: elementRef,
   });
 }
 
--- a/packages/react-resizable-panels/src/PanelGroup.ts
+++ b/packages/react-resizable-panels/src/PanelGroup.ts
@@ -109,6 +109,7 @@ export type PanelGroupProps = {
 };
 
 export type ImperativePanelGroupHandle = {
+  getElement: () => HTMLElement | null;
   getLayout: () => number[];
   setLayout: (panelSizes: number[]) => void;
 };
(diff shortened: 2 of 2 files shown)


bvaughn:
I appreciate the PR, but I would prefer not to make this change.

I don't think it actually simplifies the issue you linked to– primarily  because `PanelResizeHandle` doesn't support a ref, but you need to account for their size also. If I were to add ref support to that component, then you'd need to add multiple refs (to group and resize handle) at which point I think the code isn't really any more straight forward.

bartlangelaan:
I think it _would_ simplify the code because a ref is available on the Group. This means a queryElements can be done on this element, instead of the .document.
The original example is not scalable, because you need the reference to the element that is rendered (and not just any group rendered in the document).
One way of doing that is by adding a `<div>` around the panelgroup that you _can_ give a ref to, but I don't prefer that because it introduces a wrapper div.
It is also possible to generate an unique id, but that feels more like an escape hatch. In the end, it feels more right to me that a ref is available (just like all other React components). Hiding this feels weird.

The example could become something like:
```typescript
  const MIN_SIZE_IN_PIXELS = 100;

  const [minSize, setMinSize] = useState(10);
  const panelGroupRef = useRef<ImperativePanelGroupHandle>(null);

  useLayoutEffect(() => {
    const panelGroup = panelGroupRef.current?.getElement();
    if (!panelGroup) return;
    const resizeHandles = panelGroup.querySelectorAll(
      "[data-panel-resize-handle-id]"
    );
    const observer = new ResizeObserver(() => {
      let height = panelGroup.offsetHeight;

      resizeHandles.forEach((resizeHandle) => {
        height -= resizeHandle.offsetHeight;
      });

      // Minimum size in pixels is a percentage of the PanelGroup's height,
      // less the (fixed) height of the resize handles.
      setMinSize((MIN_SIZE_IN_PIXELS / height) * 100);
    });
    observer.observe(panelGroup);
    resizeHandles.forEach((resizeHandle) => {
      observer.observe(resizeHandle);
    });

    return () => {
      observer.disconnect();
    };
  }, []);
```

This _would_ work when multiple panelgroups are rendered in the DOM.

bvaughn:
> The original example is not scalable, because you need the reference to the element that is rendered (and not just any group rendered in the document).

I think you're overlooking a key detail in my original comment– the data id attribute.

```js
// Given a group like so
<PanelGroup direction="horizontal" id="your-id-here">

// This query selector gets you the exact same HTML element as the getElement() API you're proposing would
// Note that "your-id-here" matches to whatever prop you pass to the PanelGroup (above)
const panelGroup = document.querySelector('[data-panel-group-id="your-id-here"]');

// With it, you can query for only resize handles within the group, like so:
panelGroup.querySelectorAll("[data-panel-resize-handle-id]");
```

bartlangelaan:
But, why force this on users? 
I find it more practical to use a ref, so I don't have to mind about a component being rendered twice possibly.
A ref is always bound to the rendered element, so I don't have to worry about generating unique ids.

bvaughn:
I think I've already shared my opinion and it seems like we just disagree. I don't think your opinion is totally crazy, I just don't think this change is an improvement. :)

PR #150, opened 2023-05-20
feat: support Panel disabled

clf17222592:
When I use Conditional panels, For example:
```
<>
  <Panel>top</Panel>
  {enabled && <PanelResizeHandle />}
  {enabled ? <Panel defaultSize={40}>{content}</Panel> : <div>{content}</div>}
</>
```

When enabled changes, content will be destroyed.  The content takes a long time to render and loses state, so I don't want it to be destroyed.
In order to solve this problem, I want the Panel to provide a disabled attribute. When disabled is true, disable the resizable function of the Panel.the code becomes:
```
<>
  <Panel>top</Panel>
  {enabled && <PanelResizeHandle />}
  <Panel defaultSize={40} disabled={!enabled}>{content}</Panel>
</>
```

Or expose the PanelGroupContext so that I can customize the Panel ?

### DIFF
--- a/packages/react-resizable-panels/src/Panel.ts
+++ b/packages/react-resizable-panels/src/Panel.ts
@@ -34,6 +34,7 @@ export type PanelProps = {
   order?: number | null;
   style?: CSSProperties;
   tagName?: ElementType;
+  disabled?: boolean;
 };
 
 export type ImperativePanelHandle = {
@@ -58,6 +59,7 @@ function PanelWithForwardedRef({
   order = null,
   style: styleFromProps = {},
   tagName: Type = "div",
+  disabled = false,
 }: PanelProps & {
   forwardedRef: ForwardedRef<ImperativePanelHandle>;
 }) {
@@ -110,7 +112,7 @@ function PanelWithForwardedRef({
     }
   }
 
-  const style = getPanelStyle(panelId, defaultSize);
+  const style = disabled ? {} : getPanelStyle(panelId, defaultSize);
 
   const committedValuesRef = useRef<{
     size: number;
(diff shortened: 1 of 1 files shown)


bvaughn:
The proposed change is not one that I want to make. I think it's kind of an awkward API.

Have you tried rendering the `Panel` inside of an `unstable_Offscreen` component? That might be an option that works nicely.

Let's file an issue to continue this discussion?

clf17222592:
thank you for your reply.&nbsp;I have considered Offscreen, but it is in the experimental stage, and it is not suitable for production environments. And Offscreen can only solve the state problem, and cannot solve the problem of component re-rendering. My component is more complicated and it takes a long time to render.
This API may indeed be inelegant. Can't it also expose PanelGroupContext?




------------------&nbsp;原始邮件&nbsp;------------------
发件人: "Brian ***@***.***&gt;; 
发送时间: 2023年5月23日(星期二) 中午11:14
收件人: ***@***.***&gt;; 
抄送: "在 想 ***@***.***&gt;; ***@***.***&gt;; 
主题: Re: [bvaughn/react-resizable-panels] feat: support Panel disabled (PR #150)





 
The proposed change is not one that I want to make. I think it's kind of an awkward API.
 
Have you tried rendering the Panel inside of an unstable_Offscreen component? That might be an option that works nicely.
 
Let's file an issue to continue this discussion?

bvaughn:
That's fair, although I have been using Offscreen in production for a long time. (React DevTools uses it, as does Replay.io)

> And Offscreen can only solve the state problem, and cannot solve the problem of component re-rendering.

It sounds like your component perhaps needs to use better memoization (if rendering is so expensive).

> Can't it also expose PanelGroupContext?

I don't want to expose that context because it has internal APIs on it.

PR #151, opened 2023-05-23
lazily initialize the default storage

AntonNiklasson:
In order to avoid issues in an environment where third-party cookies are disabled, the default storage needs to be lazily initialised.

I don't have a minimal reproduction yet, I could get to that if you want me to.

### DIFF
--- a/packages/react-resizable-panels/src/PanelGroup.ts
+++ b/packages/react-resizable-panels/src/PanelGroup.ts
@@ -60,16 +60,24 @@ function throwServerError() {
   throw new Error('PanelGroup "storage" prop required for server rendering.');
 }
 
-const defaultStorage: PanelGroupStorage = {
-  getItem:
-    typeof localStorage !== "undefined"
-      ? (name: string) => localStorage.getItem(name)
-      : (throwServerError as any),
-  setItem:
-    typeof localStorage !== "undefined"
-      ? (name: string, value: string) => localStorage.setItem(name, value)
-      : (throwServerError as any),
-};
+let defaultStorage: null | PanelGroupStorage = null
+
+function initializeDefaultStorage(): PanelGroupStorage {
+  if (defaultStorage === null) {
+    defaultStorage = {
+      getItem:
+        typeof localStorage !== "undefined"
+          ? (name: string) => localStorage.getItem(name)
+          : (throwServerError as any),
+      setItem:
+        typeof localStorage !== "undefined"
+          ? (name: string, value: string) => localStorage.setItem(name, value)
+          : (throwServerError as any),
+    };
+  }
+
+  return defaultStorage
+}
 
 export type CommittedValues = {
   direction: Direction;
@@ -123,7 +131,7 @@ function PanelGroupWithForwardedRef({
   forwardedRef,
   id: idFromProps = null,
   onLayout,
-  storage = defaultStorage,
+  storage = initializeDefaultStorage(),
   style: styleFromProps = {},
   tagName: Type = "div",
 }: PanelGroupProps & {


bvaughn:
`localStorage` isn't "_initialized_". It's referenced. I think you're saying that the _referencing_ of this is causing trouble in some cases?

> I don't have a minimal reproduction yet, I could get to that if you want me to.

Repros are always helpful. Better yet, an end-to-end test would be ideal so that we could avoid a regression.

For instance, if we make a change like this– I think we may benefit from going one step further and deferring initialization until the first time storage is _actually used_ (not just during props initialization). Maybe something like?

```ts
const defaultStorage: PanelGroupStorage = {
  getItem: (name: string) => {
    if (typeof localStorage !== "undefined") {
      // Bypass this check for future calls
      defaultStorage.getItem = localStorage.getItem;
      return localStorage.getItem(name);
    } else {
      throw new Error('localStorage not supported in this environment');
    }
  },
  setItem: (name: string, value: string) => {
    if (typeof localStorage !== "undefined") {
      // Bypass this check for future calls
      defaultStorage.setItem = localStorage.setItem;
      localStorage.setItem(name, value);
    } else {
      throw new Error('localStorage not supported in this environment');
    }
  },
};
```

Haven't tested this but I think it would work?

---

PS Please revert all of the changes to `pnpm-lock.yaml`

bvaughn:
> In order to avoid issues in an environment where third-party cookies are disabled, the default storage needs to be lazily initialised.

FWIW I think you could also work around that today by just passing a no-op `storage` prop

AntonNiklasson:
> localStorage isn't "initialized". It's referenced.
I was referring to the "default storage" defined [here](https://github.com/bvaughn/react-resizable-panels/blob/main/packages/react-resizable-panels/src/PanelGroup.ts#L63-L72). Sorry about the confusion!

It seems like even just evaluating `typeof localStorage !== "undefined"` is forbidden in that kind of restricted environment. So the idea with these changes is to be able to avoid that. Previously, even when passing a custom storage, that code would run. I might be missing something here, I just want to help 😊 

A test sounds like a great idea! I understand that some kind of repro makes it easier to motivate making this change.

Thank you for building this btw! I love the minimal API, it seems really well thought out.

AntonNiklasson:
> Please revert all of the changes to pnpm-lock.yaml

Ah, not intentional! I'll get rid of that.

bvaughn:
> It seems like even just evaluating typeof localStorage !== "undefined" is forbidden in that kind of restricted environment. The idea with these changes is to be able avoid it completely.

Yeah, this makes sense 👍🏼  I think we can avoid referencing that value at all until we need to read/write.

> Thank you for building this btw! I love the minimal API, it seems really well thought out.

You're welcome! 😃

packages/react-resizable-panels/src/PanelGroup.ts
```diff
@@ -123,7 +131,7 @@ function PanelGroupWithForwardedRef({
   forwardedRef,
   id: idFromProps = null,
   onLayout,
-  storage = defaultStorage,
+  storage = initializeDefaultStorage(),
```
bvaughn, line 134: RE https://github.com/bvaughn/react-resizable-panels/pull/151#issuecomment-1559594826

> For instance, if we make a change like this– I think we may benefit from going one step further and deferring initialization until the first time storage is actually used (not just during props initialization). Maybe something like?

I think this still makes sense to do.

bvaughn:
I don't think I can do anything with this PR until there's a repro.

bvaughn:
Fixed via 5c5a4be

AntonNiklasson:
Sorry for not getting back to you, Brian. I had to tend to a lot of other things at work. Thank you for releasing a fix! 👏

PR #158, opened 2023-06-11
enable Panel callbacks when PanelGroup setLayout method is called

### LINKED ISSUE #154: Panels methods onResize and onCollapse are not triggered when setLayout is called
Hi,

On screen that I develop, I'm using the collapse feature, once I'm forcing a layout on the `PanelGroup` with the `setLayout` method,  `onResize` or `onCollapse` methods on `Panel` are not triggered 

4 Panels example: 
while being in sizes state of `[10, 20, 70, 0]`, I'm calling `setLayout([10, 0, 45, 45])`
the second Panel is not ware that he is in collapse state and the 4th Panel isn't aware that he is out from collapse state

any suggestions?

Thanx in advance

Sloebel:
fix #154

### DIFF
--- a/packages/react-resizable-panels-website/tests/Panel-OnCollapse.spec.ts
+++ b/packages/react-resizable-panels-website/tests/Panel-OnCollapse.spec.ts
@@ -18,7 +18,7 @@ async function openPage(
 
   const panelGroup = createElement(
     PanelGroup,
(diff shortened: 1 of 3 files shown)


bvaughn (approved):
Looks good 👍🏼 

I appreciate the added tests!

PR #161, opened 2023-06-22
Panel onResize not called if there is no onLayout

rijk:
For the initial render, if the layout is restored from local storage, the Panel's `onResize` function is only called if the PanelGroup has a `onLayout`. This is not expected and I assume was not intentional?

In this PR I just removed the `if (onLayout)` check in favor of a `onLayout?.(sizes)` later on.

### DIFF
--- a/packages/react-resizable-panels/src/PanelGroup.ts
+++ b/packages/react-resizable-panels/src/PanelGroup.ts
@@ -236,7 +236,9 @@ function PanelGroupWithForwardedRef({
 
     // Don't commit layout until all panels have registered and re-rendered with their actual sizes.
     if (sizes.length > 0) {
-      onLayout?.(sizes);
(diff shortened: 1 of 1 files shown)


bvaughn (approved):
Thanks!

bvaughn:
Published as 0.0.52!

bvaughn:
I think this change caused a regression, though I'm not sure how _yet_.

I tried updating to this version in Replay (https://github.com/replayio/devtools/pull/9381) and here is a recording of a failed end to end test:
https://app.replay.io/recording/affac99d-d290-4cc6-91db-6c542545ef12

rijk:
Woa, nice work tracking that down. Sorry this PR introduced that issue!

bvaughn:
No worries ☺️

PR #166, opened 2023-06-30
Add option to have transition

Mokshit06:
This PR adds an optional prop to the `Panel` component to make it possible for it to have a transition on collapse/expand. Needed for replayio/devtools#8914

### DIFF
--- a/packages/react-resizable-panels/src/Panel.ts
+++ b/packages/react-resizable-panels/src/Panel.ts
@@ -35,6 +35,7 @@ export type PanelProps = {
   order?: number | null;
   style?: CSSProperties;
   tagName?: ElementType;
+  transition?: string;
 };
 
 export type ImperativePanelHandle = {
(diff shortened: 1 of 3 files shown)


bvaughn:
I don't want to add style-related props to the `Panel` component. There are too many and it would be a rabbit hole. The component already allows custom styles to be specified via `className` or `style` props.

So basically, rather than what this PR proposes:
```tsx
<Panel transition="1s" {...props} />
```

You can already do:
```jsx
<Panel style={{
   transition: "flex-grow 1s ease-in-out"
}} {...props} />
```

Or better yet:
```jsx
<Panel className="class-with-transition" {...props} />
```

Thanks for the PR/proposal though!

Mokshit06:
Oh right, I don't know how I missed that. Thanks

bvaughn:
No worries :)

PR #167, opened 2023-07-01
Add worker condition

Andarist:
Workers (webworkers, edge workers, etc) are meant to be treated as SSR environments - they don't have access to the DOM after all. Bundlers tend to pick the `browser` condition for them (at the very least Next.js and Cloudflare's Wrangler are doing that but IIRC others are doing that as well) so we need to "override" that with an explicit `worker` condition that should be picked by them first, before picking the `browser` condition.

### DIFF
--- a/packages/react-resizable-panels/package.json
+++ b/packages/react-resizable-panels/package.json
@@ -17,11 +17,31 @@
         "import": "./dist/react-resizable-panels.cjs.mjs",
         "default": "./dist/react-resizable-panels.cjs.js"
       },
-      "development": {
-        "browser": {
+      "browser": {
+        "development": {
+          "worker": {
+            "module": "./dist/react-resizable-panels.development.worker.esm.js",
+            "import": "./dist/react-resizable-panels.development.worker.cjs.mjs",
+            "default": "./dist/react-resizable-panels.development.worker.cjs.js"
+          },
           "module": "./dist/react-resizable-panels.browser.development.esm.js",
           "import": "./dist/react-resizable-panels.browser.development.cjs.mjs",
           "default": "./dist/react-resizable-panels.browser.development.cjs.js"
+        },
+        "worker": {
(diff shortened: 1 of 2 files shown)


packages/react-resizable-panels/package.json
```diff
@@ -17,12 +17,32 @@
...
+          "worker": {
+            "module": "./dist/react-resizable-panels.development.worker.esm.js",
+            "import": "./dist/react-resizable-panels.development.worker.cjs.mjs",
+            "default": "./dist/react-resizable-panels.development.worker.cjs.js"
+          },
           "module": "./dist/react-resizable-panels.browser.development.esm.js",
           "import": "./dist/react-resizable-panels.browser.development.cjs.mjs",
           "default": "./dist/react-resizable-panels.browser.development.cjs.js"
         },
+        "worker": {
```
Andarist, line 31: This one is a little bit suboptimal. It's not incorrect - it's just that perhaps this could be somehow avoided to minimize the `exports` map? cc @emmatown 

The thing that is suboptimal is that `worker` is both nested here in `browser` and it's also used as a standalone "top-level" condition. It could be simply the top-level condition if only it would came **before** the `browser` condition.

bvaughn:
My initial reaction is that...this is getting a bit rough. ~70 lines of exports definitions in the `package.json` for a package with CJS+ESM+types. I don't see people adopting this at scale if the maintenance complexity grows like this. There is probably a limit (I assume we're near it?) with how many conditions a typical module would care about, but even so. This is a lot to manage by hand and not mess up. It's also a lot of permutations of build targets to test (and so most many of them will probably go untested, which doesn't seem great)

bvaughn:
I'm going to pass on this for now.

muqshots:
is there a plan to add this anytime soon?
Currently throws errors when running on cloudflare workers.

PR #168, opened 2023-07-01
Remove outdated comment about Parcel workaround

Andarist:
Those should no longer be needed. If you want to set those explicitly to aid React DevTools in production then I think the `as any` cast is still a preferred approach but for different reasons. TS isn't able to infer from functions with properties so some HOC-like/FP patterns don't work with `React.FC`-like types (see the thread [here](https://github.com/microsoft/TypeScript/issues/30650#issuecomment-486680485)). It's best to keep those as plain functions if possible at type-level (note that components using `forwardRef` are already breaking that rule because the return type of `forwardRef` is a function with extra properties. You can't do much about that though unless you'd be ready to use some casts with types like [this](https://github.com/adobe/react-spectrum/blob/99ca82e87ba2d7fdd54f5b49326fd242320b4b51/packages/react-aria-components/src/utils.tsx#L19-L21))

### DIFF
--- a/packages/react-resizable-panels/src/PanelContexts.ts
+++ b/packages/react-resizable-panels/src/PanelContexts.ts
@@ -17,7 +17,4 @@ export const PanelGroupContext = createContext<{
   unregisterPanel: (id: string) => void;
 } | null>(null);
 
-// Workaround for Parcel scope hoisting (which renames objects/functions).
-// Casting to :any is required to avoid corrupting the generated TypeScript types.
-// See github.com/parcel-bundler/parcel/issues/8724
-(PanelGroupContext as any).displayName = "PanelGroupContext";
+PanelGroupContext.displayName = "PanelGroupContext";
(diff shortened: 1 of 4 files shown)


packages/react-resizable-panels/src/PanelContexts.ts
```diff
@@ -16,8 +16,3 @@ export const PanelGroupContext = createContext<{
   stopDragging: () => void;
   unregisterPanel: (id: string) => void;
 } | null>(null);
-
-// Workaround for Parcel scope hoisting (which renames objects/functions).
-// Casting to :any is required to avoid corrupting the generated TypeScript types.
-// See github.com/parcel-bundler/parcel/issues/8724
-(PanelGroupContext as any).displayName = "PanelGroupContext";
```
Andarist, line 23: Actually, this one might still be needed because JS engine can't infer a name here. Context isn't a function.

bvaughn:
Most of these (not counting the context) aren't "_needed_", but I think the disambiguation in React DevTools is kind of a nice trade-off for the few extra bytes they cost.

Andarist:
Sure, the thing that caught my eye was the comment - and not the `.displayName` per se. I pushed out a change to just remove the comment.

bvaughn (approved):

PR #184, opened 2023-08-17
Distribute drags over more than just panels directly beside resizers #1

### LINKED ISSUE #181: Resizing stop working if panel to immediate right hit's it's max
https://codesandbox.io/s/react-resizable-panels-forked-lc3p4d?file=/src/App.js:663-677

If up have a panel on the left that can get smaller but it's sibling is at it's max resizing doesn't seem to work. In this situation it should probably start adding width to the next sibling with more room to grow.

https://github.com/bvaughn/react-resizable-panels/assets/1192452/c8936de6-8b17-4f20-aae6-bc49fc30c5ed

The panel in the center restricts resizing, but you can use another resizer to continue resizing

hipstersmoothie:
Took me a few tries and the code could probably be improved but this fixes #181 

The way it work now is:

- when we apply the new sizes in the direction the delta dictates instead of applying it to just that panel the code now checks if that panel is maxed. If it is we continue on to the next panel. If no panels have room to grow resizing stops
- the same logic is applied to panels in the opposite direction of the delta. we check if the panels in that direction are at their min and move on to the next panel. If there are no more panels that can shrink resizing stops.

### DIFF
--- a/packages/react-resizable-panels-website/src/routes/examples/Horizontal.tsx
+++ b/packages/react-resizable-panels-website/src/routes/examples/Horizontal.tsx
@@ -34,6 +34,7 @@ function Content() {
     <div className={styles.PanelGroupWrapper}>
       <PanelGroup className={styles.PanelGroup} direction="horizontal">
         <Panel
+          id="left"
           className={styles.PanelRow}
           defaultSize={10}
           minSize={10}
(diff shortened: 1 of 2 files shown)


bvaughn:
Hey thanks for sharing this idea 👍🏼 

It doesn't look like this was built on top of #182, so it's probably not something I would consider landing as is. (I plan to ship that branch once I have a bit more time to finish it– probably after my on call shift ends this week.) Still probably worth having a proof of concept like this to look at and consider though! :)

hipstersmoothie:
I'll update it tonight, doesn't look too different 😄

bvaughn:
#182 has been merged. Please rebase this if you want me to take a look.

bvaughn:
I'm tidying up old issues and PRs after merging the long-running refactor. I'm going to close this as it's conflicted and stale. Please re-open it after rebasing if you'd like me to review it!

PR #199, opened 2023-10-10
Adds transpile of nullish operators for webpack 4 compatibility

zamuka:


bvaughn (commented):
Added in ccce237 and will ship with the next release.

PR #201, opened 2023-10-26
VS code behaviour on open as well

kristian240:


kristian240:
@bvaughn What do you say? VSCode works like this and it would be much better UX.

bvaughn:
Once #182, I'll revisit (if this PR is still even needed) but I don't have the bandwidth to merge and rebase anything in the meanwhile.

bvaughn:
#182 has shipped! Can you rebase this and add/update tests?

bvaughn:
I think I might make this change myself as part of addressing #210

bvaughn:
Going to close this since it's against the old implementation.

PR #209, opened 2023-11-14
 fix(Panel): onResize missing prevMixedSizes on first render 

### LINKED ISSUE #208: `PanelOnResize` arg `prevMixedSizes` typed as required but is not present on init
v57
Repro: https://codesandbox.io/p/sandbox/stoic-elion-fp8m2n

Note the console output for `prevMixedSizes` is `undefined` on init

I think it makes sense to leave the type as required and make `prevMixedSizes === mixedSizes` on init. That's a nicer than typing it as optional and needing to do a lot of checks.

psychedelicious:
`Panel`'s `onResize` callback had `prevMixedSizes` undefined on first render.

- Fall back to its `mixedSizes` when it is nullish
- Add test for this case

Closes #208

### DIFF
--- a/packages/react-resizable-panels/src/utils/callPanelCallbacks.ts
+++ b/packages/react-resizable-panels/src/utils/callPanelCallbacks.ts
@@ -32,6 +32,7 @@ export function callPanelCallbacks(
     const lastNotifiedMixedSizes = panelIdToLastNotifiedMixedSizesMap[panelId];
     if (
       lastNotifiedMixedSizes == null ||
+      lastNotifiedMixedSizes == undefined ||
       mixedSizes.sizePercentage !== lastNotifiedMixedSizes.sizePercentage ||
       mixedSizes.sizePixels !== lastNotifiedMixedSizes.sizePixels
     ) {
@@ -40,7 +41,8 @@ export function callPanelCallbacks(
       const { onCollapse, onExpand, onResize } = callbacks;
 
       if (onResize) {
-        onResize(mixedSizes, lastNotifiedMixedSizes);
+        // On first render, lastNotifiedMixedSizes will be undefined, fall back to mixedSizes
+        onResize(mixedSizes, lastNotifiedMixedSizes ?? mixedSizes);
       }
 
       if (collapsible && (onCollapse || onExpand)) {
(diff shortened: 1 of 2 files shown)


psychedelicious:
Why the test is failing a bit beyond me. I'm happy to try and fix it, but I'd need some direction. May not have really fixed the issue correctly of course.

packages/react-resizable-panels/src/utils/callPanelCallbacks.ts
```diff
@@ -32,6 +32,7 @@ export function callPanelCallbacks(
     const lastNotifiedMixedSizes = panelIdToLastNotifiedMixedSizesMap[panelId];
     if (
       lastNotifiedMixedSizes == null ||
+      lastNotifiedMixedSizes == undefined ||
```
bvaughn, line 35: `== null` will match both `null` and `undefined` so this change is redundant.

packages/react-resizable-panels/src/utils/callPanelCallbacks.ts
```diff
@@ -40,7 +41,8 @@ export function callPanelCallbacks(
       const { onCollapse, onExpand, onResize } = callbacks;
 
       if (onResize) {
-        onResize(mixedSizes, lastNotifiedMixedSizes);
+        // On first render, lastNotifiedMixedSizes will be undefined, fall back to mixedSizes
+        onResize(mixedSizes, lastNotifiedMixedSizes ?? mixedSizes);
```
bvaughn, line 45: There are no "previous" sizes for the initial render, so this value _should_ be `undefined`.  I think the only thing that should be changed here is the type definition:
```ts
export type PanelOnResize = (
  mixedSizes: MixedSizes,
  prevMixedSizes: MixedSizes | undefined
) => void;
```

packages/react-resizable-panels/src/utils/callPanelCallbacks.ts
psychedelicious: TIL, ty

packages/react-resizable-panels/src/utils/callPanelCallbacks.ts
psychedelicious: Thanks, I'll close this PR and open a fresh one with the minimal type change.

psychedelicious:
Superseded by #212

PR #212, opened 2023-11-14
fix: make PanelOnResize arg prevMixedSizes optional

### LINKED ISSUE #208: `PanelOnResize` arg `prevMixedSizes` typed as required but is not present on init

psychedelicious:
It is `undefined` on first render.

Closes #208 
Context #209

FWIW, `noUncheckedIndexedAccess` in tsconfig would. have caught this, but it also catches a whole heck of a lot more.

### DIFF
--- a/packages/react-resizable-panels/src/Panel.ts
+++ b/packages/react-resizable-panels/src/Panel.ts
@@ -19,7 +19,7 @@ export type PanelOnCollapse = () => void;
 export type PanelOnExpand = () => void;
 export type PanelOnResize = (
   mixedSizes: MixedSizes,
-  prevMixedSizes: MixedSizes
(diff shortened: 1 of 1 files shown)


bvaughn (approved):

bvaughn:
Thanks!

> FWIW, `noUncheckedIndexedAccess` in tsconfig would. have caught this, but it also catches a whole heck of a lot more.

Not apposed to making this change but probably don't have time/energy to do it myself.

PR #224, opened 2023-12-04
fix: resize should not fail when `dragState` is null

### LINKED ISSUE #223: 'Cannot destructure property 'initialCursorPosition' of 'r' as it is null.' error in calculateDragOffsetPercentage
Sadly, I cannot provide clear reproduction steps, as this error occurred in our aggregated Sentry data, 73 hits during last 30 days for 7k of our internal users.

tsufiev:
Resolves #223

### DIFF
--- a/packages/react-resizable-panels/src/PanelGroup.ts
+++ b/packages/react-resizable-panels/src/PanelGroup.ts
@@ -711,17 +711,19 @@ function PanelGroupWithForwardedRef({
 
       const pivotIndices = determinePivotIndices(groupId, dragHandleId);
 
-      let delta = calculateDeltaPercentage(
-        event,
-        groupId,
-        dragHandleId,
-        direction,
-        dragState!,
-        {
-          percentage: keyboardResizeByPercentage,
-          pixels: keyboardResizeByPixels,
-        }
-      );
+      let delta = dragState
+        ? calculateDeltaPercentage(
+            event,
+            groupId,
+            dragHandleId,
+            direction,
+            dragState,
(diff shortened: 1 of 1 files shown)


tsufiev:
I have to agree that guarding against nullish `dragState` so deep in the logic is not ideal, since most like the root cause lies somewhere inside `PanelResizeHandle.ts` file. I presume that getting nullish `dragState` in the code that handles panel resizing means that `resizeHandler` has been invoked without the prior invocation of `startDragging`. However due to the overall logic of binding and unbinding event listeners inside `PanelResizeHandle.ts` being a bit convoluted, I cannot understand when and why that happens :/.

tsufiev:
@bvaughn I've managed to reproduce the issue 2 times out of ~100, so it's quite rare. Both times it happened during dragging the panel to collapsed state. The attempt of resize while `dragState` stored in `committedValuesRef` is `null` happens inside [onMove](https://github.com/bvaughn/react-resizable-panels/blob/main/packages/react-resizable-panels/src/PanelResizeHandle.ts#L112) handler. 

My mental picture of this is that when resize handle gets released and cursor is moved at the same time, sometimes the `onMove` handler gets one more chance to get executed before `useEffect` clean-up code kicks in and removes all the listeners.

bvaughn:
Cherry picking this change into my release branch caused 11 e2e tests to fail ([same as with CI](https://github.com/bvaughn/react-resizable-panels/actions/runs/7117122391/job/19491397170?pr=224)).

PR #228, opened 2023-12-07
Normalize layouts not totalling to 100%

### LINKED ISSUE #227: When total sum of persisted panel sizes differs from 100, no resize is possible
The issue can be easily reproduced by going to https://react-resizable-panels.vercel.app/examples/persistence and manually altering one of panel percentage sizes inside localStorage's value under `PanelGroup:sizes:persistence` key. 

The tricky question here is how does the app come to such a corrupted state in localStorage? Unfortunately I cannot share any insight at the moment. The only thing I know is that several users came to me complaining about inability to adjust panel sizes and the only thing I found in their browsers were those size discrepancies.

tsufiev:
Resolves #227


bvaughn:
Thanks for this! I've cherry-picked it into the upcoming release branch (#230)

PR #249, opened 2023-12-30
feat: add refs to underlying DOM elements in imperative API

### LINKED ISSUE #248: Feature Request: Access underlying DOM elements via Imperative API
## Request

With the removal of pixel-based layouts, it'd be nice to have direct access to the underlying DOM elements via the Imperative API.

For example, `panelGroupRef.current.element` could be a ref to the panel group div, and `panelRef.current.element` could be a ref to the panel div.

The specific use-case I'm thinking about is checking the size of the panel elements, to implement some pixel constraints outside the library.

## Alternatives Considered

1. Use `querySelector` to grab them based on data attributes. Awkward, requiring keeping track of however many panel & panel group IDs.
```js
document.querySelector('[data-panel-group-id="app"] [data-panel-id="gallery"]')
```

2. Wrap the panels in other elements and access them via refs. Gets messy, cluttering the DOM with divs.

psychedelicious:
Closes #248 

No worries if this feature isn't accepted, or the implementation should be different.

### DIFF
--- a/packages/react-resizable-panels/README.md
+++ b/packages/react-resizable-panels/README.md
@@ -46,7 +46,10 @@ import { Panel, PanelGroup, PanelResizeHandle } from "react-resizable-panels";
 `PanelGroup` components also expose an imperative API for manual resizing:
 | method | description
 | :-------------------------------- | :---
-| `setLayout(panelSizes: number[])` | Resize panel group to the specified _panelSizes_ (`[1 - 100, ...]`).
+| `getId(): string` | Gets the panel group's ID.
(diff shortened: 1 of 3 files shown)


bvaughn:
I don't think I want to approach exposing the DOM elements this way. Having multiple "ref" properties for a component– one that points to the DOM element and one that points to an imperative API– feels odd to me.

I have some utilities that I use internally for querying the underling DOM elements. I think maybe I can export those publicly though, to fit your original use case. They are here: https://github.com/bvaughn/react-resizable-panels/tree/main/packages/react-resizable-panels/src/utils/dom

psychedelicious:
Sounds good!

PR #252, opened 2023-12-30
feat: make element types more accurate

### LINKED ISSUE #251: Bug (ish): `tagName: ElementType` is too wide
## Problem

Exported components accept a `tagName` prop to determine which element type to use for its DOM elements.

This is currently typed as `ElementType`, but this is a bit too wide as it allows any `Element`. For example, this allows tags like `svg`, which don't have the attributes needed for the library to work and cause runtime errors. I think only `HTMLElement`s work.

It's hard to imagine a situation in which this would ever actually cause problems, but it's nice to have correct types of course.

## Proposal 1

Narrow the type to of `tagName` so that it only accepts tags for `HTMLElement`s. It looks like the way to do this is `keyof HTMLElementTagNameMap`:

```ts
export type PanelProps = Omit<HTMLAttributes<keyof HTMLElementTagNameMap>, "id" | "onResize"> &
  PropsWithChildren<{
    ...
    tagName?: keyof HTMLElementTagNameMap;
  }>;
```

I was surprised to not find a direct type e.g. `HTMLElementType`. Also, `JSX.IntrisicElements` includes non-HTML elements, so that doesn't help.

## Proposal 2

*Really* narrow the type down to just a handful of specific tags - like `div`, `span`, etc. Feels a bit restrictive, and not future-proofed.

## Proposal 3

Just update the docs with a warning.

psychedelicious:
This widens the usage of `HTMLDivElement` to `HTMLElement` throughout.

Also narrows the allowed tag names to actual HTML elements. Closes #251

### DIFF
--- a/packages/react-resizable-panels/src/Panel.ts
+++ b/packages/react-resizable-panels/src/Panel.ts
@@ -54,7 +54,7 @@ export type ImperativePanelHandle = {
   resize: (size: number) => void;
 };
 
-export type PanelProps = Omit<HTMLAttributes<ElementType>, "id" | "onResize"> &
+export type PanelProps = Omit<HTMLAttributes<keyof HTMLElementTagNameMap>, "id" | "onResize"> &
   PropsWithChildren<{
     className?: string;
     collapsedSize?: number | undefined;
@@ -68,7 +68,7 @@ export type PanelProps = Omit<HTMLAttributes<ElementType>, "id" | "onResize"> &
     onResize?: PanelOnResize;
     order?: number;
     style?: object;
-    tagName?: ElementType;
+    tagName?: keyof HTMLElementTagNameMap;
   }>;
 
 export function PanelWithForwardedRef({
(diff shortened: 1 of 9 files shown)


bvaughn:
I don't get it. It looks like you changed formatting only.

bvaughn:
Looks like this broke a bunch of things in CI.

packages/react-resizable-panels/src/Panel.ts
```diff
@@ -90,7 +93,7 @@ export function PanelWithForwardedRef({
   ...rest
 }: PanelProps & {
   forwardedRef: ForwardedRef<ImperativePanelHandle>;
-}) {
+}): ReactNode {
```
psychedelicious, line 96: When I added `keyof HTMLElementTagNameMap`, declaration generation failed despite `tsc` being happy. The error message was a red herring - the components just need an explicit return type. Not sure what to make of that.

Maybe just `ReactElement` is more appropriate?

bvaughn: That's interesting. Not sure why that is, but I think `ReactNode` makes sense as the return type 👍🏼

bvaughn (approved):
Looks good. Thanks.

Issue #260, opened 2024-01-11
Re-calculations after Conditional not behaving as expected


Psvensso:
The re-calculations after dynamically change panels don't behave as expected, however this might be fully intentional sorry for the noise if thats the case. I have not been able to narrow it down exactly to the root cause so im describing a few unexpected behaviors.

Scenario 1
1) Add collapsible={true} on the Left panel in the [Conditional example](https://github.com/bvaughn/react-resizable-panels/blob/23eb27220017c35706b323ec5cf6756ec61eee25/packages/react-resizable-panels-website/src/routes/examples/Conditional.tsx#L69)
2) Drag to collapse the Left panel
3) Click to remove the Right panel
Expected: 
The left panel should remain collapsed and the main panel should take up remaning space. 
Actuall: 
Left and Right "resets" to 50/50

Scenario 2
1) Resize Left to 50 Middle to 25 and right to 25. 
2) Hide Right panel

Expected: 
Left to take up ~2/3
Middle to take up ~1/3
Actuall: 
Panel "resets" to 50/50

Interesting observation: 
When restoring the Right to visible again the layout are restored to the "correct" previous layout. 
So without digging to deep into the re-calcuations it appears to be some "missmatch previous state" that causes it to reset the calculations instead of reusing the individual resized state by id.

Psvensso:
Observation: 
In PanelGroup.ts ~530 we try to load the state for the new panels but get a not hit (since we removed a panel), it then "re-calculates" a new state with calculateUnsafeDefaultLayout({panelDataArray,}) PanelGroup.ts:535

Perhaps we could re-use current state when calculating the new state or keep each panel state separate instead of beeing a composition.

bvaughn:
The behavior you describe in your scenarios are expected/intentional, even though I understand why it's perhaps subjective or unintuitive. Panel group stores layouts using both the `autoSaveId` (if one is provided) _and_ the combination of panels (panel attributes really). That's because the same group can be used to contain different combinations of panels/things, and persisted values are not meaningful between different combinations.

To illustrate this, let's say we have 3 panels in a group– each 33.3% width. Remove one panel and it makes no sense for the remaining two panels to both be 33% width.

[...]

PR #261, opened 2024-01-11
260 recalcresize

Psvensso:
This PR is created as a discussion/demo to describe #260 

The PR flattens the panel group state 

**From**
```typescript
{
  "group-1-key": {
    "one,two,three": {layout: [33, 33, 34], expandToSizes: {}},
    "one,two": {layout: [50, 50], expandToSizes: {}},
    "two,three": {layout: [50, 50], expandToSizes: {}},
  }
}
```
**To** 
storing each layout state in a hash
```typescript
{
  "group-1-key": {
    layout: [33, 33, 34], 
    expandToSizes: {},
    hash: {
      one: 33,
      two: 33,
      three: 34
    }
  } satisfies PanelGroupState
}
```

The layouts are then always taken from that hash so a baseline is provided when calculating the new state.
```typescript
layout = panels.map(({ id }) => state?.hash?.[id] || 0)
``` 

When then removing a panel the state might me [33, 33] but that will be recalculated. 

This is a naive implementation, i have not checked all other use-cases, for example what happends if you drag-n-drop / reorder components then this might have to be revisited.

### DIFF
--- a/packages/react-resizable-panels-website/src/routes/examples/Collapsible.tsx
+++ b/packages/react-resizable-panels-website/src/routes/examples/Collapsible.tsx
@@ -177,7 +177,7 @@ const FILES: File[] = FILE_PATHS.map(([path, code]) => {
 const CODE = `
 <PanelGroup direction="horizontal">
   <SideTabBar />
-  <Panel collapsible={true} collapsedSizePixels={35} minSize={10}>
+  <Panel collapsible={true} collapsedSize={35} minSize={10}>
     <SourceBrowser />
   </Panel>
   <PanelResizeHandle />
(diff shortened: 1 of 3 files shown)


Psvensso:
Of course the "should store layouts separately per panel combination" is exactly what im trying to remove so the tests would then not be valid.

bvaughn:
Thank you for the PR! This functionality is working as designed though, for reasons I explained in #260. Thanks for your understnding.

bvaughn:
I will still give the proposal you made some consideration, but I don't think I'll go with the change proposed in this branch. It results in unstable sizes.

Compare the behavior of the latest release...

https://github.com/bvaughn/react-resizable-panels/assets/29597/6599c0eb-8d8d-4ebc-9f4e-dbe8b955b6ab

To the behavior in this branch...

https://github.com/bvaughn/react-resizable-panels/assets/29597/dd9eb37a-2703-4953-b558-1d9589bd2961

Going to close this for now.

PR #291, opened 2024-02-10
BugFix: Resize handle hit detection respects stacking order

bvaughn:


packages/react-resizable-panels/src/PanelResizeHandleRegistry.ts
```diff
@@ -137,13 +156,39 @@ function recalculateIntersectingHandles({ x, y }: { x: number; y: number }) {
...
+        // That is why we only check potentially intersecting handles,
+        // and why we skip if the event target is within the handle's DOM
+        compare(targetElement, element) > 0 &&
+        // If the target is above the drag handle, then we also need to confirm they overlap
+        // If they are beside each other (e.g. a panel and its drag handle) then the handle is still interactive
+        intersects(
+          targetElement.getBoundingClientRect(),
+          element.getBoundingClientRect(),
+          true
+        )
```
bvaughn, line 187: Note to self:

It's not enough to compare the target here. The target might be a small element inside of a larger container. (For example, a SPAN or a DIV inside of a larger modal dialog).

websiddu:
Thanks for the hard work. Brough, you couple of coffees.

bvaughn:
I appreciate it!

PR #299, opened 2024-02-13
fix: cannot passed element id in `<Panel />` `<PanelGroup />`, `<PanelResizeHandle />`

koralle:
I found while writing  a11y tests using [axe-core](https://github.com/dequelabs/axe-core) that the `<Panel />`, `<PanelGroup />`, and `<PanelResizeHandle />` ids are not being passed properly.

### DIFF
--- a/packages/react-resizable-panels/src/Panel.test.tsx
+++ b/packages/react-resizable-panels/src/Panel.test.tsx
@@ -738,6 +738,38 @@ describe("PanelGroup", () => {
     });
   });
 
+  describe("a11y", () => {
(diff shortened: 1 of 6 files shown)


bvaughn (approved):
Thanks. This was an oversight.

bvaughn:
I think this could be considered a breaking change 🤔  I think it may be safer to only pass through ids from props (ones explicitly assigned by users, not the auto-generated ones I use internally)

bvaughn:
Thanks for the PR! This change just went out in 2.0.8

---

PR #307, opened 2024-02-23
Duplicate panel id throws AssertionError case

### LINKED ISSUE #303: Assertion Failed on conditionally rendered panels with size constraints
Hi @bvaughn, first of all thank you for this amazing library. It has been a core feature for my project. In my project, I'm using resizable panels dynamically. As in, there are a set of layouts with size constraints defined in them as well. When I switch between layouts I get an assertion error. I've found out this is due to `minSize` and `maxSize` being supplied. Most probably I'm not using them as they should so I would love it if you could shine a light on this.

Here is a repro: https://codesandbox.io/p/sandbox/youthful-jennings-wt9ffk

DenizUgur:
I've added a test case for the issue described in #303. The test case is not perfect but it throws the `AssertionError` as expected so I hope this would help to track down the problem.

**For future readers:** Ensure you have **unique** id's for your `PanelGroup`s as well, if rendering conditionally.

closes #303

### DIFF
--- a/packages/react-resizable-panels/src/PanelGroup.test.tsx
+++ b/packages/react-resizable-panels/src/PanelGroup.test.tsx
@@ -1,6 +1,6 @@
 // @ts-expect-error This is an experimental API
 // eslint-disable-next-line no-restricted-imports
-import { unstable_Activity as Activity } from "react";
+import { unstable_Activity as Activity, Fragment } from "react";
 import { Root, createRoot } from "react-dom/client";
 import { act } from "react-dom/test-utils";
 import {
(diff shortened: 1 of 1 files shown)


bvaughn:
I'm not sure what this test is showing, yet.

One thing I do notice is that– if you're dynamically rendering panels, this library asks that you specify both an `id` and an `order` prop to make sure the layouts map correctly (so when "bar" moves from being the 2nd panel, to the first, its layout is preserved). That's one of the warnings the test is surfacing.

bvaughn:
If you're swapping this group between "vertical" and "horizontal" layouts, the easiest approach here would probably be to use a **different id** for these two cases (e.g. "group-vertical" and "group-horizontal")

DenizUgur:
During my debugging I also gave a monotonically increasing number for `order` prop but that didn't do anything. Only thing that worked was ensuring outermost panel group had unique `id`.

packages/react-resizable-panels/src/PanelGroup.test.tsx
```diff
@@ -130,6 +130,110 @@ describe("PanelGroup", () => {
...
+    };
+
+    act(() => {
+      root.render(<Container container={layouts[0]} />);
+    });
+
+    act(() => {
+      root.render(<Container container={layouts[1]} />);
+    });
+  });
```
bvaughn, line 235: I _think_ this test could be reduced to the following (so that it's easier to read)?
```tsx
  it("should rerender properly after the entire layout is changed", () => {
    act(() => {
      root.render(
        <PanelGroup id="parent" direction="vertical">
          <Panel id="foo" minSize={30} order={0} />
          <PanelResizeHandle />
          <Panel id="bar" minSize={70} order={1} />
        </PanelGroup>
      );
    });

    act(() => {
      root.render(
        <PanelGroup id="parent" direction="horizontal">
          <Panel id="baz" order={0} />
          <PanelResizeHandle />
          <Panel id="bar" maxSize={25} order={1} />
        </PanelGroup>
      );
    });
  });
```
Note that I added the missing `order` prop (that was being warned about)

bvaughn:
> During my debugging I also gave a monotonically increasing number for `order` prop but that didn't do anything. Only thing that worked was ensuring outermost panel group had unique `id`.

Yes, that wouldn't fix the error you're reporting but it *is* another thing you would need to address here (and the library warns about it in DEV mode)

packages/react-resizable-panels/src/PanelGroup.test.tsx
DenizUgur: I agree, I didn't knew what was the root cause so I added my entire sandbox project. I'll revise it

bvaughn:
> And for the `order` prop, what would you do in nested groups? Each groups first panel should start with 0 or I should give an `order` in the order I render them (regardless of the group)

The order should be their rendering order (or their order in the DOM)

_Ideally_ I should be able to infer this, but it's awkward to do that with React's parent/child boundaries

packages/react-resizable-panels/src/PanelGroup.test.tsx
bvaughn: No need. I'm looking into this already. If I decide I need to add new tests to cover this area (likely) I don't mind writing them.

bvaughn: I've been able to trim it down a little farther to this failing case:
```tsx
it("should rerender properly after layout is changed", () => {
    act(() => {
      root.render(
        <PanelGroup direction="vertical">
          <Panel id="foo" minSize={30} order={0} />
          <PanelResizeHandle />
          <Panel id="bar" minSize={70} order={1} />
        </PanelGroup>
      );
    });

    act(() => {
      root.render(
        <PanelGroup direction="vertical">
          <Panel id="bar" minSize={70} order={0} />
        </PanelGroup>
      );
    });
```

bvaughn: I _think_ this change (and test) covers it
```diff
diff --git a/packages/react-resizable-panels/src/PanelGroup.test.tsx b/packages/react-resizable-panels/src/PanelGroup.test.tsx
index c6f747a..abb12ce 100644
--- a/packages/react-resizable-panels/src/PanelGroup.test.tsx
+++ b/packages/react-resizable-panels/src/PanelGroup.test.tsx
@@ -130,6 +130,35 @@ describe("PanelGroup", () => {
     expect(rightPanelElement?.getAttribute("data-panel-size")).toBe("40.0");
   });
 
+  // github.com/bvaughn/react-resizable-panels/issues/303
+  it("should recalculate layout after panels are changed", () => {
+    let mostRecentLayout: number[] | null = null;
+
+    const onLayout = (layout: number[]) => {
+      mostRecentLayout = layout;
+    };
+
+    act(() => {
+      root.render(
+        <PanelGroup direction="vertical" onLayout={onLayout}>
+          <Panel id="foo" minSize={30} order={0} />
+          <PanelResizeHandle />
+          <Panel id="bar" minSize={70} order={1} />
+        </PanelGroup>
+      );
+    });
+    expect(mostRecentLayout).toEqual([30, 70]);
+
+    act(() => {
+      root.render(
+        <PanelGroup direction="vertical" onLayout={onLayout}>
+          <Panel id="bar" minSize={70} order={0} />
+        </PanelGroup>
+      );
+    });
+    expect(mostRecentLayout).toEqual([100]);
+  });
+
   describe("imperative handle API", () => {
     it("should report the most recently rendered group id", () => {
       const ref = createRef<ImperativePanelGroupHandle>();
diff --git a/packages/react-resizable-panels/src/PanelGroup.ts b/packages/react-resizable-panels/src/PanelGroup.ts
index 469d0e6..ce167c7 100644
--- a/packages/react-resizable-panels/src/PanelGroup.ts
+++ b/packages/react-resizable-panels/src/PanelGroup.ts
@@ -768,10 +768,10 @@ function PanelGroupWithForwardedRef({
         panelData,
         layout
       );
-      assert(
-        prevPanelSize != null,
-        `Previous panel size not found for panel "${panelData.id}"`
-      );
+      if (prevPanelSize == null) {
+        // It's possible that the panels in this group have changed since the last render
+        return;
+      }
 
       if (
         prevCollapsible &&
```

bvaughn:
Thanks for sharing this repro. I think it's helped me track down a bug and a gap in my unit test coverage. I think I have a fix in #309 so I'm going to close this PR.

PR #313, opened 2024-03-06
BugFix: Global cursor style reset on unmount

### LINKED ISSUE #308: Global cursor style persists if PanelResizeHandle is disabled while dragging
I'm trying to implement a vertical split pane view in an application, and am copying code over from another frontend app our company is developing that does something similar using an older version of this library.

In both applications, the bottom panel collapses if it's brought close to the bottom of the screen, and the panel resize handle is disabled with the intention being for the user to click one of the buttons in the footer to open it again.

In the original application, using v0.0.55 of the library, this works without issues. However, in my application, using v2.0.9 of the library, I find that sometimes the cursor gets stuck with the `ns-resize` appearance through the whole page, as shown below: 

https://github.com/bvaughn/react-resizable-panels/assets/1588951/9407bcce-b51f-4c1a-8672-66ba53a1a2f4

I've determined that it's related to this code: https://github.com/bvaughn/react-resizable-panels/blob/e922a963a90f530cb294eaf4bfd1450fa1246d27/packages/react-resizable-panels/src/utils/cursor.ts

What seems to be happening is that there's a race condition between the global cursor style getting reset when the mouse button is released, and the PanelResizeHandle getting disabled when the cursor is released near the bottom of the screen. If the latter event handler "wins", then it seems to interrupt the event handler that resets the global cursor style.

I think a potential fix might be to update the following code to ensure the global cursor style is reset if the PanelResizeHandle is disabled: 

https://github.com/bvaughn/react-resizable-panels/blob/e922a963a90f530cb294eaf4bfd1450fa1246d27/packages/react-resizable-panels/src/PanelResizeHandle.ts#L104C1-L111C71

TAGC:
Fixes #308.

This PR updates the `PanelResizeHandle` component to call `resetGlobalCursorStyle` when unmounted. This should prevent the cursor from indefinitely retaining the "resize" style if the component is unmounted as a panel is being resized.

I've tested out these changes locally and they appear to resolve the issue.

### DIFF
--- a/packages/react-resizable-panels/src/PanelResizeHandle.test.tsx
+++ b/packages/react-resizable-panels/src/PanelResizeHandle.test.tsx
@@ -49,12 +49,12 @@ describe("PanelResizeHandle", () => {
   });
 
   afterEach(() => {
-    jest.clearAllMocks();
-    jest.resetModules();
-
     act(() => {
       root.unmount();
     });
+
+    jest.clearAllMocks();
+    jest.resetModules();
 
     expect(expectedWarnings).toHaveLength(0);
   });
@@ -92,6 +92,8 @@ describe("PanelResizeHandle", () => {
         </PanelGroup>
       );
     });
+
+    expect(cursorUtils.resetGlobalCursorStyle).not.toHaveBeenCalled();
 
     act(() => {
       root.unmount();
--- a/packages/react-resizable-panels/src/PanelResizeHandle.test.tsx
+++ b/packages/react-resizable-panels/src/PanelResizeHandle.test.tsx
@@ -16,8 +16,9 @@ import {
 import * as cursorUtils from "./utils/cursor";
 
 jest.mock("./utils/cursor", () => ({
-  ...jest.requireActual("./utils/cursor"),
+  getCursorStyle: jest.fn(),
   resetGlobalCursorStyle: jest.fn(),
+  setGlobalCursorStyle: jest.fn(),
 }));
 
 describe("PanelResizeHandle", () => {
@@ -80,26 +81,6 @@ describe("PanelResizeHandle", () => {
     expect(element.tabIndex).toBe(123);
     expect(element.getAttribute("data-test-name")).toBe("foo");
     expect(element.title).toBe("bar");
-  });
-
-  it("resets the global cursor style on unmount", () => {
-    act(() => {
-      root.render(
-        <PanelGroup direction="horizontal">
-          <Panel />
-          <PanelResizeHandle />
-          <Panel />
-        </PanelGroup>
-      );
-    });
-
-    expect(cursorUtils.resetGlobalCursorStyle).not.toHaveBeenCalled();
-
-    act(() => {
-      root.unmount();
-    });
-
-    expect(cursorUtils.resetGlobalCursorStyle).toHaveBeenCalled();
   });
 
   function setupMockedGroup({
@@ -287,7 +268,7 @@ describe("PanelResizeHandle", () => {
       act(() => {
         leftElement.focus();
       });
-      expect(document.activeElement).toBe(leftElement);
+      // expect(document.activeElement).toBe(leftElement);
       verifyAttribute(leftElement, "data-resize-handle-active", "keyboard");
       verifyAttribute(rightElement, "data-resize-handle-active", null);
 
@@ -334,5 +315,34 @@ describe("PanelResizeHandle", () => {
       expect(element).not.toBeNull();
       expect(element?.getAttribute("id")).toBeNull();
     });
+  });
+
+  fit("resets the global cursor style on unmount", () => {
+    const onDraggingLeft = jest.fn();
+
+    const { leftElement } = setupMockedGroup({
+      leftProps: { onDragging: onDraggingLeft },
+      rightProps: {},
+    });
+
+    act(() => {
+      dispatchPointerEvent("pointermove", leftElement);
+    });
+
+    act(() => {
+      dispatchPointerEvent("pointerdown", leftElement);
+    });
+    expect(onDraggingLeft).toHaveBeenCalledTimes(1);
+    expect(onDraggingLeft).toHaveBeenCalledWith(true);
+
+    expect(cursorUtils.resetGlobalCursorStyle).not.toHaveBeenCalled();
+    expect(cursorUtils.setGlobalCursorStyle).toHaveBeenCalledTimes(1);
+
+    act(() => {
+      root.unmount();
+    });
+
+    expect(cursorUtils.resetGlobalCursorStyle).toHaveBeenCalled();
+    expect(cursorUtils.setGlobalCursorStyle).toHaveBeenCalledTimes(1);
   });
 });
--- a/packages/react-resizable-panels/src/PanelResizeHandle.test.tsx
+++ b/packages/react-resizable-panels/src/PanelResizeHandle.test.tsx
@@ -7,13 +7,13 @@ import {
   type PanelResizeHandleProps,
 } from ".";
 import { assert } from "./utils/assert";
+import * as cursorUtils from "./utils/cursor";
 import { getResizeHandleElement } from "./utils/dom/getResizeHandleElement";
 import {
   dispatchPointerEvent,
   mockBoundingClientRect,
   verifyAttribute,
 } from "./utils/test-utils";
-import * as cursorUtils from "./utils/cursor";
 
 jest.mock("./utils/cursor", () => ({
   getCursorStyle: jest.fn(),
@@ -50,12 +50,12 @@ describe("PanelResizeHandle", () => {
   });
 
   afterEach(() => {
+    jest.clearAllMocks();
+    jest.resetModules();
+
     act(() => {
       root.unmount();
     });
-
-    jest.clearAllMocks();
-    jest.resetModules();
 
     expect(expectedWarnings).toHaveLength(0);
   });
@@ -268,7 +268,7 @@ describe("PanelResizeHandle", () => {
       act(() => {
         leftElement.focus();
       });
-      // expect(document.activeElement).toBe(leftElement);
+      expect(document.activeElement).toBe(leftElement);
       verifyAttribute(leftElement, "data-resize-handle-active", "keyboard");
       verifyAttribute(rightElement, "data-resize-handle-active", null);
 
@@ -317,7 +317,7 @@ describe("PanelResizeHandle", () => {
     });
   });
 
-  fit("resets the global cursor style on unmount", () => {
+  it("resets the global cursor style on unmount", () => {
     const onDraggingLeft = jest.fn();
 
     const { leftElement } = setupMockedGroup({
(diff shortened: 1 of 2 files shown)


packages/react-resizable-panels/src/PanelResizeHandle.test.tsx
```diff
@@ -1,14 +1,24 @@
 import { Root, createRoot } from "react-dom/client";
 import { act } from "react-dom/test-utils";
-import type { PanelResizeHandleProps } from "react-resizable-panels";
-import { Panel, PanelGroup, PanelResizeHandle } from ".";
+import {
+  Panel,
+  PanelGroup,
+  PanelResizeHandle,
+  type PanelResizeHandleProps,
+} from ".";
```
TAGC, line 8: I've updated the way `PanelResizeHandleProps` gets imported because the original way doesn't seem to work for me, even after a clean checkout. I can revert this if necessary.

TAGC: That fixes it, thanks.

packages/react-resizable-panels/src/PanelResizeHandle.ts
```diff
@@ -101,9 +102,17 @@ export function PanelResizeHandle({
     committedValuesRef.current.state = state;
   });
 
+  // Reset cursor style on unmount
+  useEffect(() => {
+    return () => {
+      resetGlobalCursorStyle();
+    };
+  }, []);
```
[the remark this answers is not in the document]
bvaughn: I think the right fix might be- on unregister- if there are any current intersecting drag handles, re-compute them and then update the cursor again (using the same methods already defined in that module)

TAGC:
Just an update on my end - work on this has been deprioritised as the team for the other frontend have found a workaround for the issue they were experiencing, and I've had higher priority stuff come through. I'd still like to continue on this when I get a chance, though.

bvaughn:
Thanks for the update!

packages/react-resizable-panels/src/PanelResizeHandleRegistry.ts
```diff
@@ -73,9 +73,48 @@ export function registerResizeHandle(
     if (count === 1) {
       ownerDocumentCounts.delete(ownerDocument);
     }
+
+    recalculateIntersectingHandlesAfterUnregister(element);
   };
 }
 
+function recalculateIntersectingHandlesAfterUnregister(
```
TAGC, line 81: @bvaughn 

These changes were written by ChatGPT after I explained the problem to it and fed it the source code for the relevant files. It seems to work as expected; I've not yet observed the issue re-appear in the project I have linked against the forked version of this library. It seems to match your recommended approach:

> I think the right fix might be- on unregister- if there are any current intersecting drag handles, re-compute them and then update the cursor again (using the same methods already defined in that module)

I asked ChatGPT (and had a look myself) whether it would be possible to call the existing `recalculateIntersectingHandles` function within the unregister callback, but it seems this would require more refactoring and potentially be more brittle than the solution ChatGPT proposes that involves writing these two new functions. We can discuss whether there's a way to simplify this further and re-use more of the existing code, though.

TAGC:
Hi @bvaughn, would you be able to re-review this when you get a chance? It's not critical, but it would be useful for my hackday project.

packages/react-resizable-panels/src/PanelResizeHandle.test.tsx
```diff
@@ -72,6 +82,24 @@ describe("PanelResizeHandle", () => {
...
+      root.render(
+        <PanelGroup direction="horizontal">
+          <Panel />
+          <PanelResizeHandle />
+          <Panel />
+        </PanelGroup>
+      );
+    });
+
+    act(() => {
```
bvaughn, line 96: This test could give a false positive, as it's currently written. You can fix this by verifying that the mock function hasn't been called before you expected.
```suggestion
    });

    expect(cursorUtils.resetGlobalCursorStyle).not.toHaveBeenCalled();

    act(() => {
```

packages/react-resizable-panels/src/PanelResizeHandleRegistry.ts
bvaughn: > These changes were written by ChatGPT

I remember why I didn't re-review this. This was kind of a turn off to me. In my experience, ChatGPT generates code that looks plausible but is often wrong in som way.

In this case, I don't think this new function is necessary. It's called by `unregisterResizeHandle` **after** the unmounting handle has been removed from `registeredResizeHandlers` (so that special case check is not needed). Given that, wouldn't the `recalculateIntersectingHandles` method work for updating the intersecting handles?

We'd still want to call `updateCursor` afterward but it will already call `resetGlobalCursorStyle` if there are no intersecting handles.

packages/react-resizable-panels/src/PanelResizeHandleRegistry.ts
```diff
@@ -73,9 +73,48 @@ export function registerResizeHandle(
...
+  });
+
+  if (intersectingHandles.length > 0) {
+    updateCursor();
+  } else {
+    resetGlobalCursorStyle();
+  }
+}
+
+function checkIfHandleIntersects(data: ResizeHandlerData): boolean {
```
bvaughn, line 102: This function is also unnecessary, as `recalculateIntersectingHandles` already has logic for this.

bvaughn (changes requested):

packages/react-resizable-panels/src/PanelResizeHandleRegistry.ts
TAGC: > I remember why I didn't re-review this. This was kind of a turn off to me. In my experience, ChatGPT generates code that looks plausible but is often wrong in som way.

No worries, that's understandable and the reason I suspected.

I'd been looking over this code for a while without figuring out what exactly I needed to write, so I made use of ChatGPT to give me a starting point. I wanted to be forthright about it being AI-generated code for exactly the reasons you mention, but in this case I know it's generated something along the right lines as I've been able to verify its solution through unit testing and linking the local version of the library against my project.

You're right that there are some parts of the solution that didn't look right, even to me, such as the following:

```ts
if (intersectingHandles.length > 0) {
  updateCursor();
} else {
  resetGlobalCursorStyle();
}
```

Since we know that `resetGlobalCursorStyle` will be called by `updateCursor` if necessary anyway. However, I wanted to give ChatGPT's solution "as-is" to start with, since I don't understand the codebase as well as you and any amendments to its code on my part might be worse than the amendments you would have made. For example, in this case I would probably try to amend this code to be:

```ts
if (intersectingHandles.length == 0) {
  resetGlobalCursorStyle();
}
```

But I'm not 100% confident that's the best solution here.

What I'll try to do is incorporate the feedback you've given on ChatGPT's "as-is" solution to produce something simpler that looks right, still passes the unit test (now updated with your suggested), and still fixes the issue when linked against my project.

packages/react-resizable-panels/src/PanelResizeHandle.test.tsx
TAGC: Good shout, I've made that change.

bvaughn (commented):
I do not feel comfortable landing this code as-is without addressing the unnecessary functions

packages/react-resizable-panels/src/PanelResizeHandleRegistry.ts
```diff
@@ -73,6 +73,10 @@ export function registerResizeHandle(
     if (count === 1) {
       ownerDocumentCounts.delete(ownerDocument);
     }
+
+    const dummyEvent = { target: null, x: 0, y: 0 };
+    recalculateIntersectingHandles(dummyEvent);
+    updateCursor();
```
TAGC, line 79: Would you prefer I pass a dummy event to `recalculateIntersectingHandles` as done here or update the function itself to make this parameter optional?

packages/react-resizable-panels/src/PanelResizeHandleRegistry.ts
TAGC: > Given that, wouldn't the recalculateIntersectingHandles method work for updating the intersecting handles?

It was a while ago but I remember now - my original plan was to try exactly this, but `recalculateIntersectingHandles` expects an argument of the form:

```ts
{
  target: EventTarget | null;
  x: number;
  y: number;
}
```

Elsewhere, this function is invoked in response to an event e.g. `handlePointerDown`, `handlePointerMove`, etc. There was no event data to pass it from `unregisterResizeHandle` so I wasn't sure how I'd actually make use of it.

Based on your feedback, I've had another look at it and this time just tried passing a set of dummy event data. This passes the unit test and appears to work fine in my project, and limits the production code changes to just three additional lines.

I really appreciate your patience with all this, by the way.

bvaughn: Hey no worries. And sorry for the slow response. Work was very busy this week and I'm just circling back to this.

The fake event question is a good one. I think we would actually need the mouse's most recent (current) position for that approach work. (It's likely that nothing would intersect with a fake 0,0 coordinate, so it would effectively always remove the cursor.) We'd really want it to recalculate and only remove the cursor if no remaining (still mounted) panels intersect with the mouse.

Maybe there's a better way to approach this. We keep the most recent set of intersecting handles in `intersectingHandles`. Maybe we should just remove the unmounting handle from that list, and then based on the remaining set– update the cursor. (This would handle the case where the set was empty and re removed the cursor entirely _as well as_ the case where it went from e.g. 2 handles with a "+" cursor to only one.)

Does this make sense? Sorry this only got half of my attention last week so I probably sent you down the wrong path.

packages/react-resizable-panels/src/PanelResizeHandle.test.tsx
```diff
@@ -72,6 +82,26 @@ describe("PanelResizeHandle", () => {
...
+
+    expect(cursorUtils.resetGlobalCursorStyle).not.toHaveBeenCalled();
+
+    act(() => {
+      root.unmount();
+    });
+
+    expect(cursorUtils.resetGlobalCursorStyle).toHaveBeenCalled();
+  });
+
```
bvaughn, line 104: This test should probably be one of the end to end tests. Right now it's not _really_ testing the right behavior (since the cursor is never set in the first place, because there are no intersecting handles in this test)

bvaughn:

packages/react-resizable-panels/src/PanelResizeHandleRegistry.ts
```diff
@@ -73,6 +73,17 @@ export function registerResizeHandle(
...
+    // If the resize handle that is currently unmounting is intersecting with the pointer,
+    // update the global pointer to account for the change
+    if (intersectingHandles.includes(data)) {
+      const index = intersectingHandles.indexOf(data);
+      if (index >= 0) {
+        intersectingHandles.splice(index, 1);
+      }
+
+      updateCursor();
+    }
```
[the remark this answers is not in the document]
TAGC: That seems to make sense, yeah.

TAGC:
Thanks for taking this forward. I've updated my project to use v2.0.20 of this package and it's looking good to me :)

PR #317, opened 2024-03-14
fix: parent cycle problem

fundon:


bvaughn:
Thank you!

bvaughn:
react-resizable-panels@2.0.13

Issue #336, opened 2024-04-12
Elements behind hitAreaMargins receive mouse events inconsistently


4nnikaf:
**The general problem**
Elements that are behind the hitAreaMargins of a PanelResizeHandle receive mouse events inconsistently.

`onClick` is blocked by the hitAreaMargin and is not triggered on the element beneath it.
`onMouseEnter/onMouseLeave` events however are NOT blocked, creating hover effects that might suggest to the user that the element was clickable.
`onPointerDown` is NOT blocked by the hitAreaMargins. So depending on the implementation clicks may be triggered on the element if the onPointerDown event is consumed instead of the onClick event.
`onPointerUp` however does get blocked.

I made this codesandbox to show the described issue: https://codesandbox.io/p/sandbox/react-resize-panels-event-issue-forked-m74dnk

**Desired behavior**
All pointer events are treated equally and get blocked by the hitAreaMargin. Or there is at least a way to make it behave this way. I currently couldn't figure out a way to prevent these events from going through to the panel content.
Generally if the cursor turns into the resize cursor and the user clicks and drags to resize the panel, that is the only thing that should be happening.

**My specific use case**
I am currently implementing a component where you can move and split panels via Drag&Drop, like you're used to from any IDE, and I'm using dnd-kit for the Drag&Drop. dnd-kit listens to the onPointerDown event, which is still triggered underneath the hitAreaMargins. So when I start resizing a panel on the hitAreaMargins, dnd-kit also starts dragging the panel at the same time, which make the entire component pretty much unusable. I struggle to find a way to prevent the onPointerDown event from being triggered for dnd-kit, when the resize handle's margin is the actual target for the action.
I tried setting the hitAreaMargins to 0, but there is still a thin line where the event clearly says that it's target is the dnd-kit drag handle, but the PanelResizeHandle still thinks it's being dragged. Though I could not reproduce that issue in the codesandbox, so maybe that's just a funky interaction between the libraries.

**Probably a minor side issue**
If I click on the hitAreaMargin while hovering the element beath it, the onMouseLeave event is called, even though the cursor is still hovering the element. I'd exepect the mouse leave/enter events to either function normally or to get fully blocked by the hitAreaMargins.
This can also be seen in the codesandbox example. I'm not sure if this is the same problem, so I could create a separate issue for this, though it's not currently a problem I'm having in my use case.

PR #338, opened 2024-04-15
Prevent click event from triggering on elements behind hitAreaMargins

4nnikaf:
Fixing a problem described in #336 

To solve the issue of click events passing through the hitAreaMargins to the elements behind them all we need to do is handle the events in the capture phase and prevent them from bubbling down to the children/panel content by calling `event.stopPropagation()`.

However cancelling the mouse/touch events (for exmaple `mousedown`/`touchstart`), does not cancel their pointer event counterparts (`pointerdown`), like it does the other way around. So I also changed the event handlers to listen to the pointer events instead.

### DIFF
--- a/packages/react-resizable-panels/src/PanelResizeHandleRegistry.ts
+++ b/packages/react-resizable-panels/src/PanelResizeHandleRegistry.ts
@@ -89,6 +89,7 @@ function handlePointerDown(event: ResizeEvent) {
     updateResizeHandlerStates("down", event);
 
     event.preventDefault();
+    event.stopPropagation();
   }
 }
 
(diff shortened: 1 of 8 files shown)


packages/react-resizable-panels/src/utils/events/getResizeEventCoordinates.ts
```diff
@@ -1,20 +1,19 @@
...
 export function getResizeEventCoordinates(event: ResizeEvent) {
-  if (isMouseEvent(event)) {
+  if (isPointerEvent(event)) {
+    if (event.isPrimary) {
+      return {
+        x: event.clientX,
+        y: event.clientY,
+      };
+    }
+  } else if (isMouseEvent(event)) {
```
bvaughn, line 12: This is a nice simplification

packages/react-resizable-panels/src/PanelResizeHandleRegistry.ts
```diff
@@ -89,6 +89,7 @@ function handlePointerDown(event: ResizeEvent) {
     updateResizeHandlerStates("down", event);
 
     event.preventDefault();
+    event.stopPropagation();
```
bvaughn, line 92: This is the only change that gives me pause. I guess it's okay.

bvaughn (approved):

bvaughn:
Thanks for the PR. I'm preparing a release in #339.

packages/react-resizable-panels/src/PanelResizeHandleRegistry.ts
bvaughn: This part has proven controversial (see #467)

I am considering backing the change out and requiring that "_elements behind hitAreaMargins_" check `event.defaultPrevented` to decide whether they should respond to an event or not, but I don't know how cumbersome that would be for application code to integrate with.

I would appreciate you weighing in on this if you have any thoughts, @4nnikaf

PR #352, opened 2024-05-03
Update `@replayio/playwright`

Andarist:


bvaughn:
Oh thanks. Didn't see this. Already pushed an update that does this.

PR #353, opened 2024-05-08
Update `@replayio/playwright`

Andarist:
This one makes sure that the `@replayio/reply` (transitive dependency) gets updated too (u can see the change in the lockfile).

With this one, I can record this repo and I get recordings. Some fail at times when the metadata's size gets exceeded. We know about the issue and we are in the process of figuring out the solution.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
   "devDependencies": {
     "@babel/preset-typescript": "^7.22.5",
     "@playwright/test": "^1.37.0",
-    "@replayio/playwright": "3.0.0-alpha.7",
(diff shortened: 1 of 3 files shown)


bvaughn:
Failed:

https://github.com/bvaughn/react-resizable-panels/actions/runs/9003551222/job/24734429716?pr=353

Screenshot 2024-05-08 at 10 04 15 PM

Andarist:
FWIW - the errors above were caused by missing API key on the fork. The alpha version of the plugin didn't benefit from the recent improvements to error reporting in this situation (those were just merged to `main`).

PR #348, opened 2024-05-01
Record Playwright tests with Replay

bvaughn:

PR #356, opened 2024-05-10
Use a nicer `list` reporter locally

Andarist:
This addresses the complaint from https://github.com/bvaughn/react-resizable-panels/pull/348#issuecomment-2102774377 . I didn't realize that the default is `'list'` locally, I added `'line'` reporter since I've seen that being used in a couple of other projects. I totally agree that `'list'` is quite nice and I prefer it to `'line'` - especially if we consider smaller test suites.

⚠️ `@replayio/playwright` still spits out a bunch of junk into the stdout. That makes the current output very cluttered - even with this improvement. This is going to be addressed with PRO-315

### DIFF
--- a/packages/react-resizable-panels-website/playwright.config.ts
+++ b/packages/react-resizable-panels-website/playwright.config.ts
@@ -19,8 +19,10 @@ const config: PlaywrightTestConfig = {
           apiKey: process.env.REPLAY_API_KEY,
           upload: true,
         })
-      : ["line"],
-  ],
+      : undefined,
+    // replicating Playwright's defaults
+    process.env.CI ? (["dot"] as const) : (["list"] as const),
+  ].filter((v): v is NonNullable<typeof v> => !!v),
   use: {
     browserName: "chromium",
     headless: true,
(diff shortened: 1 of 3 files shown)


packages/react-resizable-panels-website/playwright.config.ts
```diff
@@ -19,8 +19,10 @@ const config: PlaywrightTestConfig = {
           apiKey: process.env.REPLAY_API_KEY,
           upload: true,
         })
-      : ["line"],
-  ],
+      : undefined,
+    // replicating Playwright's defaults
+    process.env.CI ? (["dot"] as const) : (["list"] as const),
+  ].filter((v): v is NonNullable<typeof v> => !!v),
```
Andarist, line 25: with the upcoming TS 5.5 you will be able to rewrite this as:
```suggestion
  ].filter(v => !!v),
```

bvaughn: Nice

bvaughn (approved):

PR #366, opened 2024-06-24
Fixes TS annotation issue

Akkuma:
This fixes `The inferred type of 'ResizableHandle' cannot be named without a reference to 'react-resizable-panels/dist/declarations/src/PanelResizeHandleRegistry'. This is likely not portable. A type annotation is necessary.`

I've exported out types not currently exported from `index.ts`. This resolves the issue that there were internal types not exported resolving annotation issues for `PanelResizeHandle`. I created a build to verify this as well https://www.npmjs.com/package/@akkuma/react-resizable-panels.

### DIFF
--- a/packages/react-resizable-panels/src/index.ts
+++ b/packages/react-resizable-panels/src/index.ts
@@ -29,9 +29,7 @@ import type {
   PanelResizeHandleOnDragging,
   PanelResizeHandleProps,
 } from "./PanelResizeHandle";
-import type {
-  PointerHitAreaMargins,
-} from "./PanelResizeHandleRegistry";
+import type { PointerHitAreaMargins } from "./PanelResizeHandleRegistry";
 
 export {
   // TypeScript types
--- a/packages/react-resizable-panels/src/index.ts
+++ b/packages/react-resizable-panels/src/index.ts
@@ -29,6 +29,13 @@ import type {
   PanelResizeHandleOnDragging,
   PanelResizeHandleProps,
 } from "./PanelResizeHandle";
+import type {
+  PointerHitAreaMargins,
+  ResizeHandlerAction,
+  ResizeHandlerData,
+  SetResizeHandlerState,
+} from "./PanelResizeHandleRegistry";
+import type { Direction, ResizeEvent, ResizeHandler } from "./types";
 
 export {
   // TypeScript types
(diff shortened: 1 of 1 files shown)


Akkuma:
> > This fixes `The inferred type of 'ResizableHandle' cannot be named without a reference to 'react-resizable-panels/dist/declarations/src/PanelResizeHandleRegistry'. This is likely not portable. A type annotation is necessary.`
> 
> Where does this error come from?

This error is in an internal package that I built at my company that uses shadcn, which uses this package (https://ui.shadcn.com/docs/components/resizable). Our internal package uses vite to bundle up our customized version:

```export default defineConfig({
	publicDir: './src/css',
	build: {
		lib: {
			entry: Object.fromEntries(entries),
			formats: ['es'],
			fileName: '[name]',
		},
		rollupOptions: {
			external: ['react', 'react-dom', 'react/jsx-runtime', 'tailwindcss', 'tailwindcss-animate'],
		},
		target: 'esnext',
		sourcemap: true,
		emptyOutDir: true,
	},
	plugins: [
		muteWarningsPlugin(warningsToIgnore),
		react(),
		dts({ bundledPackages: ['class-variance-authority'], tsconfigPath: './tsconfig.build.json' }),
	],
});
```

`dts` is `vite-plugin-dts` and is likely raising this error preventing the generation of a `.d.ts` as some of the exports I added were private. So you could say that the source of this is taking this package -> packaging it up in our package. If we had no intermediary package I'd expect no issues in consuming this all directly.

> The types you've added exports for don't seem like types that should be exported.

I thought it made sense to just expose all of the types at the same time, but I could pair this back too.

bvaughn:
Thanks for the extra context. that's helpful.

What would be the minimal set of types needing to be exported to fix the error you are referencing?

Akkuma:
> Thanks for the extra context. that's helpful.
> 
> What would be the minimal set of types needing to be exported to fix the error you are referencing?

I quickly checked and  that would be:
```ts
import type { PointerHitAreaMargins } from "./PanelResizeHandleRegistry.js";
export { PointerHitAreaMargins }
```
I believe the reason for this is that `PointerHitAreaMargins` is part of the prop for `PanelResizeHandle` and being not exported makes it "not portable" in the extra hop downstream.

bvaughn:
Interesting. Okay.

Can you update this PR to just make the minimal change?

bvaughn (approved):
Thanks

bvaughn:
CI is complaining about formatting. Mind running `pnpm run prettier`

Akkuma:
> CI is complaining about formatting. Mind running `pnpm run prettier`

Got it 👍

bvaughn:

PR #369, opened 2024-07-02
Do not call onResize on mount if the defaultSize is provided

myklt:
Actual behavior: `onResize` should be called only, when the actual change happens
Expected behavior: `onResize` should not be called on mount, when the passed size matches the provided default size
Workaround: check on user's side, if the default size is provided and `prevSize` is `undefined`. 
Proposed solution: do the check on the library side

### DIFF
--- a/packages/react-resizable-panels/src/Panel.test.tsx
+++ b/packages/react-resizable-panels/src/Panel.test.tsx
@@ -779,7 +779,7 @@ describe("PanelGroup", () => {
     });
 
     describe("onResize", () => {
-      it("should be called on mount", () => {
+      it("should be called on mount when default size is not provided", () => {
         let onResizeLeft = jest.fn();
         let onResizeMiddle = jest.fn();
         let onResizeRight = jest.fn();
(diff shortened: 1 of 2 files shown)


bvaughn:
This change seems reasonable and I understand why you would propose it. I'm concerned that it makes the library's behavior less predictable though. In my experience, you sometimes need to write code that syncs a value to external state (e.g. `onResize(value => setState(value)`) and it's nicer in that case when the callback fires for the initial value also. With your proposal, you could just initialize state to the same value as `defaultSize` but that can be something you forget to do (especially when the code might be in two different locations, like a custom hook). The current behavior protects against that, at the cost of an extra state update– but that update should be low cost, since React would bail out of any potential render work if the same value was passed as was already in the state. Does that make sense?

bvaughn:
That makes sense!

In the case of external side effects like you describe, it seems to me like this change could complicate the implementation too, but that would depend on whether the side effect cared about the initial/default size. In my experience with other similar component libraries, they often do, which makes the syncing logic a bit more complicated. (You need to write a mount effect to send the initial value, and then the rest can be sent in by the change handler.)

To me there doesn't seem to be a clear "right way" to do this. Given that, I'll pass on this suggested change. (It would be a breaking change and require a major release, and I'm not convinced it's better just equally valid.)

Thanks for the discussion!

PR #370, opened 2024-07-02
call onFocus/onBlur

hipstersmoothie:
Noticed the component accepted the props but didn't use them


bvaughn:
Hmm seems like this change broke one of the `PanelResizeHandle` unit tests. Not sure why yet.

bvaughn:

hipstersmoothie:
Thanks!

bvaughn:
Yup!

PR #373, opened 2024-07-17
fix: still resizes after releasing mouse outside window (#340)

### LINKED ISSUE #340: Still resizes after releasing mouse outside window
Check out the sandbox: https://codesandbox.io/p/sandbox/react-resizable-panels-forked-r6zcm6

If you drag the resizer up until the first panel gets collapsed, and you release the mouse outside the iframe of codesandbox (or just outside your browser window) your code does not recognize the mouse release and still changes the panel size after the mouse hovers back over the iframe.

in-ch:
Resolves #340 

Hello. 
Added features to prevent the panel size from changing when the mouse cursor is out of the browser.

### DIFF
--- a/packages/react-resizable-panels/src/PanelResizeHandle.test.tsx
+++ b/packages/react-resizable-panels/src/PanelResizeHandle.test.tsx
@@ -279,6 +279,18 @@ describe("PanelResizeHandle", () => {
       verifyAttribute(leftElement, "data-resize-handle-active", null);
       verifyAttribute(rightElement, "data-resize-handle-active", null);
     });
+
+    it("should stop dragging when the mouse goes beyond the browser size", () => {
+      const { leftElement, rightElement } = setupMockedGroup();
+      verifyAttribute(leftElement, "data-resize-handle-active", null);
+      verifyAttribute(rightElement, "data-resize-handle-active", null);
+
+      act(() => {
+        dispatchPointerEvent("pointermove", leftElement, { clientX: -1 });
+      });
+      verifyAttribute(leftElement, "data-resize-handle-state", "inactive");
+      verifyAttribute(rightElement, "data-resize-handle-state", "inactive");
+    });
   });
 
   describe("a11y", () => {
--- a/packages/react-resizable-panels/src/PanelResizeHandle.ts
+++ b/packages/react-resizable-panels/src/PanelResizeHandle.ts
@@ -193,6 +193,28 @@ export function PanelResizeHandle({
     stopDragging,
   ]);
 
+  useEffect(() => {
+    const handleMouseOut = (event: MouseEvent) => {
+      if (
+        event.clientX < 0 ||
+        event.clientY < 0 ||
+        event.clientX > window.innerWidth ||
+        event.clientY > window.innerHeight
+      ) {
+        setState("hover");
+        stopDragging();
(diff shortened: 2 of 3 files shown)


packages/react-resizable-panels/src/PanelResizeHandle.test.tsx
```diff
@@ -279,6 +279,18 @@ describe("PanelResizeHandle", () => {
...
+      const { leftElement, rightElement } = setupMockedGroup();
+      verifyAttribute(leftElement, "data-resize-handle-active", null);
+      verifyAttribute(rightElement, "data-resize-handle-active", null);
+
+      act(() => {
+        dispatchPointerEvent("pointermove", leftElement, { clientX: -1 });
+      });
+      verifyAttribute(leftElement, "data-resize-handle-state", "inactive");
+      verifyAttribute(rightElement, "data-resize-handle-state", "inactive");
+    });
```
bvaughn, line 293: This test seems to give a false positive. (I can remove the change you made entirely and it will still pass.)

bvaughn: FWIW I also think this fix may make the most sense in the resize handle registry (where other Window event listeners are) but I haven't thought about it _much_

bvaughn: The reason the test gives a false positive is that the handle is not _down_ when the move event is fired. I will take a pass at fixing the test.

bvaughn: Sorry, I forgot to leave an update. As I was looking at this the other day, I realize that I already had code in place to handle this (in the resize handler registry) and it mostly seems to work, but it doesn't work in the case of Code Sandbox (probably something specific to iframes seemingly not receiving "pointerleave" events). Might be worth focusing on that use case.

bvaughn: It occurs to me that my e2e tests need to include a route that uses an iframe at the root. I'll add one.

bvaughn: Interestingly, having just added an iframe– this library seems to work fine. So I'm not sure what it is about Code Sandbox that's different. 🤔

bvaughn: Oh! I think I just realized what's going on by accidentally pushing a reference to `localhost` in my commit. I think the issue is when the iframe is in a different domain than the parent frame. Lemme poke at that a bit more.

bvaughn: Okay, I'm able to trigger a similar behavior by using `sandbox="allow-scripts"`. I'll move forward from there and try to add a failing e2e test.

bvaughn: Check out #374; I think I found a fix.

bvaughn:
Thank you for the contribution. I ended up fixing this in #374 after looking into it.

bvaughn:
Fixed in #374 and released in 2.0.21

---

in-ch:
Thank you so much 😀

bvaughn:
You are welcome! Thanks for the coffee! 😌

PR #382, opened 2024-07-31
Further obfuscate reference to React.useId

### LINKED ISSUE #381: export 'useId' was not found in 'react'
Projects using this package and building with React <18 (the `useId` hook was introduced in React 18) and `esbuild` v0.19.6 or new (v0.23.0 is the latest version) see an error similar to the following:

```
export 'useId' (imported as 'Du') was not found in 'react' (possible exports: Children, Component, Fragment, Profiler, PureComponent, StrictMode, Suspense, __SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED, cloneElement, createContext, createElement, createFactory, createRef, forwardRef, isValidElement, lazy, memo, useCallback, useContext, useDebugValue, useEffect, useImperativeHandle, useLayoutEffect, useMemo, useReducer, useRef, useState, version)
```

It seems like starting with `esbuild` v0.19.6, the line [here](https://github.com/bvaughn/react-resizable-panels/blob/main/packages/react-resizable-panels/src/vendor/react.ts#L40C1-L40C66) is minified from:

```typescript
const useId = (React as any)["useId".toString()] as () => string;
```

to just:

```typescript
const useId = React.useId;
```

and therefore is no longer sufficiently obfuscated. I believe this issue similar in nature to #82.  It appears [other projects facing similar issues](https://github.com/mui/material-ui/issues/41190#issuecomment-2040873379) have further obfuscated the reference to `React.useId` to:

```typescript
const useId = (React as any)[`useId${Math.random()}`.slice(0, 5)] as () => string;
```

You can see this in action using the `esbuild` playground link below.  Using the Version pulldown in the top-right you can see how the old version was sufficiently obfuscated up until `esbuild` v0.19.6, where `esbuild` apparently became smart enough to minify it to just `React.useId`.

https://esbuild.github.io/try/#dAAwLjIzLjAALS1taW5pZnkALy8gY3VycmVudCB2ZXJzaW9uLCBtaW5pZmllcyB0byAiUmVhY3QudXNlSWQiClJlYWN0WyJ1c2VJZCIudG9TdHJpbmcoKV07CgovLyBwcm9wb3NlZCB2ZXJzaW9uLCBtaW5pZmllcyB0byAiUmVhY3RbYHVzZUlkJHtNYXRoLnJhbmRvbSgpfWAuc2xpY2UoMCw1KV0iClJlYWN0W2B1c2VJZCR7TWF0aC5yYW5kb20oKX1gLnNsaWNlKDAsIDUpXTs

Would you be opposed to further obfuscating this line using the `Math.random()` technique above?  If so I would be happy to open a PR with the change.  Thanks!

nwidger:
Use technique from this comment: https://github.com/mui/material-ui/issues/41190#issuecomment-2040873379 to further obfuscate reference to `React.useId` which causes issue when using `react-resizable-panels` with React <18 and `esbuild` > 0.19.5.

Fixes #381

### DIFF
--- a/packages/react-resizable-panels/src/vendor/react.ts
+++ b/packages/react-resizable-panels/src/vendor/react.ts
@@ -36,8 +36,8 @@ const {
   useState,
 } = React;
 
-// `toString()` prevents bundlers from trying to `import { useId } from 'react'`
-const useId = (React as any)["useId".toString()] as () => string;
+// `Math.random()` and `.slice(0, 5)` prevents bundlers from trying to `import { useId } from 'react'`
+const useId = (React as any)[`useId${Math.random()}`.slice(0, 5)] as () => string;
 
 const useLayoutEffect_do_not_use_directly = useLayoutEffect;
 
--- a/packages/react-resizable-panels/src/vendor/react.ts
+++ b/packages/react-resizable-panels/src/vendor/react.ts
@@ -37,7 +37,9 @@ const {
 } = React;
 
 // `Math.random()` and `.slice(0, 5)` prevents bundlers from trying to `import { useId } from 'react'`
-const useId = (React as any)[`useId${Math.random()}`.slice(0, 5)] as () => string;
+const useId = (React as any)[
+  `useId${Math.random()}`.slice(0, 5)
+] as () => string;
 
 const useLayoutEffect_do_not_use_directly = useLayoutEffect;
 


nwidger:
@bvaughn Any thoughts on the proposed changes?

packages/react-resizable-panels/src/vendor/react.ts
```diff
@@ -36,8 +36,8 @@ const {
   useState,
 } = React;
 
-// `toString()` prevents bundlers from trying to `import { useId } from 'react'`
-const useId = (React as any)["useId".toString()] as () => string;
+// `Math.random()` and `.slice(0, 5)` prevents bundlers from trying to `import { useId } from 'react'`
+const useId = (React as any)[`useId${Math.random()}`.slice(0, 5)] as () => string;
```
bvaughn, line 40: I think the formatter wants this:
```suggestion
const useId = (React as any)[
  `useId${Math.random()}`.slice(0, 5)
] as () => string;
```

bvaughn:
Hey @nwidger. Busy week for me (lots of interviews) so I hadn't looked at this yet.

Looks okay I guess. JS bundlers are _weird_. I wish there was a standard directive or something that we could use to opt out of this kind of "smart" optimization but 🤷🏼 

There's a formatting issue that's causing CI to be unhappy. I left a comment.

bvaughn:
Sorry it's taking me a while to get back to this. Been a really busy few days on my side.

PR #385, opened 2024-08-11
Fix: Resolved startup issue for Example site on Windows environment

benishouga:
Updated the build and watch npm scripts in the package.json of the react-resizable-panels-website package to make it work in a Windows environment. These changes were based on the package.json from the react-resizable-panels package.

It would be great if it also works on other environments like Mac.


bvaughn (approved):

bvaughn:
> It would be great if it also works on other environments like Mac.

Since I'm a Mac user, and the main developer here, that's definitely a requirement 😄 but this change is fine.

PR #386, opened 2024-08-14
Add opt-in support for setting the "nonce" attribute on the global cursor style

### LINKED ISSUE #384: Add support for CSP nonce
Can you please consider adding support for a `nonce` attribute to the `style` element that this library injects in the DOM. The custom cursor logic currently doesn't work when CSP is configured to expect a specific nonce.

References:
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/nonce
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/style-src

bvaughn:
Resolves #384

```js
import { setNonce } from "react-resizable-panels";

setNonce("whatever");
```

cc @asual


bvaughn:
Published in version 2.1.0

---

PR #410, opened 2024-09-23
Before stack comparison, check if target element is in document

### LINKED ISSUE #409: Stacking Order Error for Unrelated events
After looking at the source code, it seems like for handle calculation, the library attaches a `pointerup` listener to the window rather than the handle. This causes an issue with clicking on disappearing elements (for instance, ShadCN UI's select element) where every so often, clicking on a disappearing element will cause an unhandled runtime error because the nonexistent element and the registered panels don't share an ancestor.

To reproduce, have an existing panel group and use the ShadCN Select component (click on a few items a bunch of times, this error doesn't happen every time)

`"react-resizable-panels": "^2.0.23",`

exoRift:
Fixes #409 

Don't feel pressured to merge, but I figured that the solution is fairly simple so I implemented it myself.

### DIFF
--- a/packages/react-resizable-panels/src/PanelResizeHandleRegistry.ts
+++ b/packages/react-resizable-panels/src/PanelResizeHandleRegistry.ts
@@ -197,6 +197,7 @@ function recalculateIntersectingHandles({
       // and the element that was actually clicked/touched
       if (
         targetElement !== null &&
+        document.contains(targetElement) &&
         dragHandleElement !== targetElement &&
         !dragHandleElement.contains(targetElement) &&
         !targetElement.contains(dragHandleElement) &&


bvaughn:
Yeah no worries. I appreciate it. It's late where I'm at but I'll leave this open until I have time to look at it a little closer. If it doesn't seem like it will cause any negative impact, I don't mind landing it.

exoRift:
Any updates? 👀

bvaughn:
Truthfully I have not thought about it much yet. Week days are busy (job) and this is not a huge priority for me.

bvaughn:
I don't think it should break or affect any pre-existing apps so I think it's safe to make. Appreciate your patience.

bvaughn:
This change has been released in react-resizable-panels@2.1.4

---

Issue #407, opened 2024-09-19
Issue with DragEventHandler<keyof HTMLElementTagNameMap> Type


in-ch:
Hello. 😀

When using the `Panel` component with event handlers such as `onDragLeave`, `onDragEnter`, `onDrop`, and `onDragOver` typed as `DragEventHandler<keyof HTMLElementTagNameMap>`, a type error occurs when trying to use the `contains` method, which is a method of `HTMLElement`.

Here's an example where the issue occurs:
```tsx
const handleDragLeave: DragEventHandler<keyof HTMLElementTagNameMap> = (
  e: DragEvent<keyof HTMLElementTagNameMap>,
): void => {
  if (e.currentTarget.contains(e.relatedTarget)) return;
  setIsDragActive(false);
};
```

![스크린샷 2024-09-12 12 20 00](https://github.com/user-attachments/assets/1653f7b8-5975-40b5-9f80-a36cab2d7700)

The issue is that the `keyof HTMLElementTagNameMap` type represents string keys for tag names, so it cannot infer the types of methods (e.g., contains) that the event object has. 

Therefore, the `DragEventHandler<keyof HTMLElementTagNameMap>` type is limited in using methods of HTMLElement.

How about using a `union type` or something similar to make it compatible with both `keyof HTMLElementTagNameMap` and `HTMLElement types`?

bvaughn:
> How about using a union type or something similar to make it compatible with both keyof HTMLElementTagNameMap and HTMLElement types?

I don't think this fixes the typing situation you describe (at least not when I just did a quick check locally). Want to share a TypeScript REPL or something?

in-ch:
Thank you for your response. 😀

I am sharing the [CodeSandbox link](https://codesandbox.io/p/sandbox/issue-with-drageventhandler-keyof-htmlelementtagnamemap-type-xmj8nd?file=%2Fsrc%2FuseDrag.ts) for part of the code used in the project.

In the `useDrag` hook, I defined the `handleDragLeave` function using the type `React.DragEvent<HTMLDivElement>` in order to use the `currentTarget` value from the event object. 

However, when I wrote the code this way, I encountered the following error in the `onDragLeave` function of the <Panel> component:

```bash
Type '(e: React.DragEvent<HTMLDivElement>) => void' is not assignable to type 'DragEventHandler<keyof HTMLElementTagNameMap>'.
  Types of parameters 'e' and 'event' are incompatible.
[...]

PR #414, opened 2024-09-27
Add flexible tagName support to PanelProps

in-ch:
related to (#407)

Instead of fixing the `tagName` type to `keyof HTMLElementTagNameMap`, it has been made flexible by using a generic type.
This allows TypeScript to better infer `event` object types like handleDrop when a specific tagName is provided in the component.

```jsx
<Panel
    onDrop={(e: DragEvent<HTMLDivElement>) => {
      if (e.currentTarget.contains(e.relatedTarget)) return;
    }}
    className={styles.PanelRow}
    defaultSize={20}
    minSize={10}
    tagName="div"
  >
    <div className={styles.Centered}>left</div>
</Panel>

```

### DIFF
--- a/packages/react-resizable-panels/src/Panel.ts
+++ b/packages/react-resizable-panels/src/Panel.ts
@@ -54,10 +54,9 @@ export type ImperativePanelHandle = {
   resize: (size: number) => void;
 };
 
-export type PanelProps = Omit<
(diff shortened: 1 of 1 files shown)


bvaughn:
I like this approach!

PR #418, opened 2024-10-01
Add 'hover' state Tailwind class using data attribute

CHE1RON:
Resize action is visible before hitting the resize handle, so cursor change is triggered while `hover:bg-xxx` not. This makes sure the UI changes resize handle background as soon as the cursor shows "resizable"

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ This likely means that you haven't applied any CSS to style the resize handles. 
 
 ```tsx
 // Tailwind example
-<PanelResizeHandle className="w-2 bg-blue-800" />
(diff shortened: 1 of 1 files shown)


bvaughn:
I'm not sure I want to encourage this type of data-attribute based styling. The data attributes primarily exist as a way for me to programmatically test the resize panel. I don't think I want to encourage people to think of them as public API.

CHE1RON:
I see, but given the current setup, it seems logical to make both transitions at the same time: changing background color and showing "resize" cursor 🤔

PR #422, opened 2024-10-22
feat: add react v19 to peer deps

Yonom:
Thanks a lot for this package

I get the following warning from pnpm after upgrading to NextJS v15 (released today), which installs react 19 RC

React 19 is technically still in RC stage, but since a stable version of a major Metaframework is installing it, it's a sign that it's ready for production use and will now be used by many projects.

```ts
└─┬ react-resizable-panels 2.1.4
  ├── ✕ unmet peer react@"^16.14.0 || ^17.0.0 || ^18.0.0": found 19.0.0-rc-65a56d0e-20241020
  └── ✕ unmet peer react-dom@"^16.14.0 || ^17.0.0 || ^18.0.0": found 19.0.0-rc-65a56d0e-20241020
```


packages/react-resizable-panels/package.json
```diff
@@ -81,8 +81,8 @@
     "react-dom": "experimental"
   },
   "peerDependencies": {
-    "react": "^16.14.0 || ^17.0.0 || ^18.0.0",
-    "react-dom": "^16.14.0 || ^17.0.0 || ^18.0.0"
+    "react": "^16.14.0 || ^17.0.0 || ^18.0.0 || ^19.0.0 || ^19.0.0-rc",
+    "react-dom": "^16.14.0 || ^17.0.0 || ^18.0.0 || ^19.0.0 || ^19.0.0-rc"
```
bvaughn, line 85: I _think_ the `^19.0.0-rc` bit may not be necessary. Is it?

Yonom: If you want to support NextJS 15 which currently installs the RC, then it is necessary

You can test: https://semver.npmjs.com/

Screenshot 2024-10-22 at 15 58 40

bvaughn: That's interesting. I was testing using [this tool](https://jubianchi.github.io/semver-check/#/^19.0.0/19.0.0-rc-123) which seemed to suggest that it was unnecessary. I guess it's safer to use the more conservative approach though

bvaughn (approved):

bvaughn:
Thanks for the PR~

packages/react-resizable-panels/package.json
Yonom: @bvaughn  The other data point is radix-ui using the same, see:
https://github.com/radix-ui/primitives/blob/main/packages/react/dialog/package.json#L49

I am not sure about the way the package managers handle this, but generally 19.0.0 is considered higher than 19.0.0-rc.x, so if you specify "^19.0.0", you wouldn't want 19.0.0-rc (which was released before 19.0.0) to be accepted

bvaughn:
Dependencies updated in react-resizable-panels@2.1.5

---

PR #423, opened 2024-10-26
Used dir prop of PanelGroup for rtl support instead of document.dir

mohas:
This allow us to use this component in scenarios where PanelGroup is used as a component in an application where you don't have access to document.dir in order to change it for any reason, there is already a dir prop as it is standard HTMLElement prop we just have to use it instead of relying on document.dir which is not a good practice at all, one may choose to apply dir at any level below document to an element and its children, for example:

* story book component
* components with dir=rtl and ltr side by side in a single page
* ...

thanks for awesome library

### DIFF
--- a/packages/react-resizable-panels/src/PanelGroup.ts
+++ b/packages/react-resizable-panels/src/PanelGroup.ts
@@ -648,9 +648,10 @@ function PanelGroupWithForwardedRef({
 
       // Support RTL layouts
       const isHorizontal = direction === "horizontal";
-      if (document.dir === "rtl" && isHorizontal) {
+      const isElmentsRtl = (rest.dir ?? document.dir) === "rtl";
+      if (isHorizontal && isElmentsRtl) {
         delta = -delta;
-      }
+      };
 
       const panelConstraints = panelDataArray.map(
         (panelData) => panelData.constraints


bvaughn:
I'm open to the idea of allow rtl/ltr to be specified on the component level, and I understand why you'd suggest using the [`dir` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/dir) for this, but I think it complicates the DevX. TypeScript won't provide any hints about this value, but more importantly– having both `direction` and `dir` seems confusing.

We could address the first issue by updating the `PanelGroupProps` props definition to explicitly call out `dir?: "auto" | "ltr" | "rtl"` but I'm concerned about the awkward mix of `dir` and `direction`.

The only alternative approach that comes to mind would be to overload the `direction` prop to enable this (e.g. `"direction" | "direction-ltr" | "direction-rtl"`)

What are your thoughts here?

mohas:
OK, thanks for the response, here are my thoughts:

1. relying on document.anything is not recommended practice overall, as this is a component and should not assume access to outside of it's boundaries
2. `dir` is standard html attribute and by specifying `HTMLAttributes<keyof HTMLElementTagNameMap>` as part of the type for `PanelGroup` people expect dir to be there and work, the autocomplete also show dir to people
3. many design system and component libraries support providing a `dir` to their subtree through context api, for example in `antd` or `radixui` which shadcn is based on, the user base of these libraries are non-trivial, adopting the same practice would simplify things as lots of people are used to this, this change will likely be adopted by these ui libraries as for example shadcn uses this package
4. in libraries there is usually an `orientation: 'horizontal' | 'vertical'` instead of `direction`, direction is commonly used for writing direction in css
5. the changes should be in a way that does not break code that use this package and already rely on direction and document.dir hack
6. I would introduce an `orientation?: 'horizontal' | 'vertical'` prop that works exactly as the current `direction` and mark the direction as obsolete so people would gradually replace it, and use `dir` for rtl and ltr as it is already there, just not implemented

if you agree I would be happy to change this PR

bvaughn:
> `dir` is standard html attribute and by specifying `HTMLAttributes<keyof HTMLElementTagNameMap>` as part of the type for `PanelGroup` people expect dir to be there and work, the autocomplete also show dir to people

Yes, but it wouldn't suggest valid _values_ for it.

> I would introduce an `orientation?: 'horizontal' | 'vertical'` prop that works exactly as the current `direction` and mark the direction as obsolete so people would gradually replace it, and use `dir` for rtl and ltr as it is already there, just not implemented

I know we're getting into semantics now, but I think of "orientation" as being something related to a screen or a device (rotation, maybe aspect ration). "direction" in this case is the correct term, I think.

> relying on document.anything is not recommended practice overall, as this is a component and should not assume access to outside of it's boundaries

Agreed that this library shouldn't be reading this value from the document. I think this is the most compelling point in the thread here, and we shouldn't be relying on either the `document` or requiring an explicit attribute. We should probably be using `getComputedStyle` to read the inherited direction style (which I think will work with both CSS `direction` and the `dir` attribute (based on my own testing).

I'll take a pass at this and tag you on a PR.

bvaughn:
The change described above has been published in react-resizable-panels@2.1.6

PR #424, opened 2024-10-27
Don't read "direction" from the document; use inherited style instead

bvaughn:
cc @mohas


bvaughn:
Ideally I would add some unit tests for this but JSDom doesn't support either `dir` nor inherited styles, so the tests wouldn't be very valuable.

bvaughn:
Released in react-resizable-panels@2.1.6

PR #427, opened 2024-11-04
feat(Support SVG elements on modals): expand utils for stacking order to include SVGElements

gatzjames:
The issue:
When hovering over SVG icons on elements "above" a handle (e.g. modal) the resize cursor appears and users can drag it to resize the panels "bellow".

Suggested solution:
Expand the `recalculateIntersectingHandles` function to include `SVGElement` when looking at the stacking order of elements which fire pointer events.


bvaughn:
Thanks! Sorry I didn't notice this.

bvaughn:
Published as react-resizable-panels@2.1.7

---

PR #434, opened 2024-11-17
Add note about `autoSaveId`

jordiup:
hi @bvaughn,

i spent a couple hours debugging panel size persistence,

only to discover that i was using `id` instead of `autoSaveId`

it would be good if this was more clear in the readme

cheers :)

great library

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,10 @@ This likely means that you haven't applied any CSS to style the resize handles. 
 <PanelResizeHandle className="w-2 bg-blue-800" />
 ```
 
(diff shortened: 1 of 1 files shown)


bvaughn:
I'm sorry to hear that you lost time debugging this. I don't think it should be the _first_ FAQ question (based on how often it has come up) but I guess adding it seems reasonable. It would need to be done in more than just the file you modified though, so I've just done it myself in fa58630e0ba09070f6977ce0ab036588a8749ce4. Cheers

PR #435, opened 2024-11-17
Move package to ESM + ESLint 9

ahkhanjani:
In this PR I have:
- Made the npm package to use ESM internally (to support ESLint 9).
- Updated ESLint dependencies to latest and added some rules to be more strict. Removing the outdated `no-restricted-imports` plugin caught a problem on its own.
- Fixed the lint errors.


ahkhanjani:
@bvaughn Is this PR ok with you?

E2E tests are failing because of commonjs tests. So how about we drop cjs entirely and release a v3? I can do it if it's ok with you

packages/react-resizable-panels/src/Panel.test.tsx
```diff
@@ -907,8 +905,8 @@ describe("PanelGroup", () => {
     });
 
     it("should support sizes with many decimal places", () => {
-      let panelRef = createRef<ImperativePanelHandle>();
-      let onResize = jest.fn();
+      const onResize = jest.fn();
+      const panelRef = createRef<ImperativePanelHandle>();
```
bvaughn, line 909: TBH I don't really like this ESLint rule.

packages/react-resizable-panels/src/Panel.ts
```diff
@@ -243,7 +243,7 @@ export function PanelWithForwardedRef({
 
     // CSS selectors
     "data-panel": "",
-    "data-panel-collapsible": collapsible || undefined,
+    "data-panel-collapsible": collapsible ?? undefined,
```
bvaughn, line 246: This would be a backwards breaking change for styling.

The way the rule was written before was intentional. It avoids adding the data attribute _unless_ the value is true.

(`?? undefined` would be redundant.)

packages/react-resizable-panels/src/PanelResizeHandle.test.tsx
```diff
@@ -27,7 +27,7 @@ describe("PanelResizeHandle", () => {
   let container: HTMLElement;
 
   beforeEach(() => {
-    // @ts-expect-error
+    // @ts-expect-error globalThis is not typed
```
bvaughn, line 30: This comment seems kind of confusing. The code isn't reference `globalThis`.

I'm also kind of curious why every `global.IS_REACT_ACT_ENVIRONMENT` comment is slightly different in this PR 😆

packages/react-resizable-panels/src/hooks/useIsomorphicEffect.ts
```diff
@@ -3,6 +3,6 @@ import { useLayoutEffect_do_not_use_directly } from "../vendor/react";
 
 const useIsomorphicLayoutEffect = isBrowser
   ? useLayoutEffect_do_not_use_directly
-  : () => {};
+  : () => undefined;
```
bvaughn, line 6: This change seems unexpected/unnecessary to me. Did ESLint actually complain about it? 🤔

packages/react-resizable-panels/src/hooks/useUniqueId.ts
```diff
@@ -10,7 +10,7 @@ export default function useUniqueId(
 ): string {
   const idFromUseId = wrappedUseId();
 
-  const idRef = useRef<string | null>(idFromParams || idFromUseId || null);
+  const idRef = useRef<string | null>(idFromParams ?? idFromUseId ?? null);
```
bvaughn, line 13: Avoiding nullish coalescing operators was intentional (see 8677dab742347e533c1b499d3256e55c32950277)

bvaughn (commented):
I appreciate the effort spent on this PR. It's kind of a noisy change though, and (as it is) I think it could cause a couple of small regressions (commented). I'm not sure if I want to move forward with it.

packages/react-resizable-panels/src/PanelResizeHandle.test.tsx
ahkhanjani: > I'm also kind of curious why every `global.IS_REACT_ACT_ENVIRONMENT` comment is slightly different in this PR 😆

Since this is opinionated I actually wanted your review on what I should write for these (I also may or may not have let AI write those for me 😅)

packages/react-resizable-panels/src/Panel.test.tsx
bvaughn: I don't _specifically_ prefer `let` over `const` but I also don't think ESLint should have an opinion on the matter. Especially for something like a React ref, which is mutable, declaring it as a "constant" is kind of misleading– because it can still be _changed_ (mutated).

packages/react-resizable-panels/src/hooks/useIsomorphicEffect.ts
bvaughn: > An arrow function that returns undefined instead of an empty function body (that may seem like it returns an object)

FWIW `() => {}` does not return anything. It just defines an empty function, same as `function() {}`

If you wanted to return an object from an inline function, you'd need to wrap with parens `() => ({})`

ahkhanjani: > FWIW `() => {}` does not return anything. It just defines an empty function, same as `function() {}`
> 
> If you wanted to return an object from an inline function, you'd need to wrap with parens `() => ({})`

Yes. I was trying to say it looks similar to returning an object.

Should we keep this version or revert?

packages/react-resizable-panels/src/Panel.test.tsx
ahkhanjani: > I don't _specifically_ prefer `let` over `const` but I also don't think ESLint should have an opinion on the matter. Especially for something like a React ref, which is mutable, declaring it as a "constant" is kind of misleading– because it can still be _changed_ (mutated).

I think using `const` makes the code more readable (because we won't assume the variable is getting assigned to later on). It's not really about mutability and more about direct assignment. So like in the code when I see `let strict = false` I will just know that the value of `strict` is being assigned directly later in the code. Whereas in something like an object (e.g. React ref) we can't know it's being mutated regardless of `let` or `const`. I personally think being strict about that is a good thing.

I disabled the rule regardless 😁

ahkhanjani: Should I also revert the `let` to `const` changes?

ahkhanjani: > I think using `const` makes the code more readable (because we won't assume the variable is getting assigned to later on). It's not really about mutability and more about direct assignment. So like in the code when I see `let strict = false` I will just know that the value of `strict` is being assigned directly later in the code. Whereas in something like an object (e.g. React ref) we can't know it's being mutated regardless of `let` or `const`. I personally think being strict about that is a good thing.

Lines 64 to 67 of `Panel.test.tsx` is a good example of this.

```ts
describe("collapse and expand", () => {
  let leftPanelRef = createRef<ImperativePanelHandle>();
  let rightPanelRef = createRef<ImperativePanelHandle>();
  
  let mostRecentLayout: number[] | null;
  ...
}
```

Since they have `let` it will give you a good context of what's going on later on in the code. Without the ESLint rule strictly fighting against this it's impossible to know that though.

bvaughn: > Without the ESLint rule strictly fighting against this it's impossible to know that though

The entire test is 25 lines long. It’s pretty quick to just look at where the variables are referenced to see how they’re being used.

I understand your preference, and I don’t think it’s crazy. I just don’t share it 🙂 and I’m the only one who maintains this project so I think it’s reasonable to stick with my own preferences.

ahkhanjani:
@bvaughn What should we do about the cjs issue? Should we drop cjs support and go for a v3?

packages/react-resizable-panels/src/Panel.test.tsx
bvaughn: Oh no worries at all. I hope my comment didn't seem harsh.

My stance here is pretty subjective and I just wanted to clarify that.

> I'm just trying to help a library I like be maintained

I really appreciate this 🙇🏼

packages/react-resizable-panels/src/hooks/useIsomorphicEffect.ts
bvaughn: I don't feel strongly either way on this one. It just seemed unexpected.

bvaughn:
Hmm. I don't think there's a strong reason to drop CJS support, unless it would in some way dramatically improve the source code/maintainence?

ahkhanjani:
Hey @bvaughn. I fixed an import issue I had made and now one test is strangely failing:

```
PanelGroup › invariants › should throw if default size is less than 0 or greater than 100

Expected substring: "Invalid layout"                                                                                                               
Received message:   ""
```

The test itself makes sense but I couldn't find anywhere in the code that throws this exception. Why wasn't the error failing before?

bvaughn:
Going to close this PR in favor of #474 and #476 but I appreciate the work you did here 🙇🏼

PR #462, opened 2025-04-15
Use group+panel ID for panel HTML element ID to avoid duplicate eleme…

levimeahan:
In cases where a custom ID is passed for a panel (commonly done when conditionally rendering) the documentation suggests the ID only has to be unique within a group, and shows an example with a fixed string `id` value.

This causes invalid markup if the same ID is used in two groups, for instance if the `PanelGroup` is used in a shared component that renders in two places.

image

Attaching the group ID here will ensure we write unique IDs to the DOM.

### DIFF
--- a/packages/react-resizable-panels/src/Panel.ts
+++ b/packages/react-resizable-panels/src/Panel.ts
@@ -235,7 +235,7 @@ export function PanelWithForwardedRef({
 
     children,
     className: classNameFromProps,
-    id: idFromProps,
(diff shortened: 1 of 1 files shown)


bvaughn:
This would be a breaking change for any code that referenced these ids (e.g `document.getElementById("...")`)

I'd welcome a PR that updates the documentation to make this constraint clearer, but I don't think I can merge this change as is.

PR #463, opened 2025-04-18
Pass auto-generated Panel id to `id` attribute to fix `aria-controls` prop

### LINKED ISSUE #450: Auto-generated values for aria-* are invalid
When loading up [an example](https://react-resizable-panels.vercel.app/examples/horizontal), and running Lighthouse to check accessibility, the generated HTML has (what I believe to be) invalid values for selectors and therefore raises an accessibility error.

```html
<div class="AzW8qW_ResizeHandle " role="separator" tabindex="0" data-panel-group-direction="horizontal" data-panel-group-id=":r0:" data-resize-handle="" data-resize-handle-state="inactive" data-panel-resize-handle-enabled="true" data-panel-resize-handle-id=":r2:" aria-controls=":r1:" aria-valuemax="50" aria-valuemin="20" aria-valuenow="30" style="touch-action: none; user-select: none;"></div>
```

Unfortunately, the Lighthouse error isn't very specific to state which value is invalid. However, the issue stems from `:` character being not allowed, for example, in `aria-controls=":r1:"` This is proven when setting the `id` attribute manually on `Panel` to a valid selector, which also makes the Lighthouse error go away.

Image

bvaughn:
Resolves #450

### DIFF
--- a/packages/react-resizable-panels-website/index.html
+++ b/packages/react-resizable-panels-website/index.html
@@ -1,13 +1,11 @@
 <!doctype html>
-<html>
-  <html>
-    <head>
-      <title>React components for resizable panels</title>
-      <meta name="viewport" content="width=device-width, initial-scale=1" />
-      <link rel="stylesheet" href="./root.css" />
-      <link rel="stylesheet" href="./CodeMirror.css" />
-    </head>
-  </html>
+<html lang="en">
+  <head>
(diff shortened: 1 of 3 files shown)


bvaughn:
Fix has been published in react-resizable-panels@2.1.8

---

PR #465, opened 2025-04-21
Add onPointerDown, onPointerUp, and onClick callbacks to PanelResizeHandle

### LINKED ISSUE #438: Detect click events inside hit area
First, thank you for the library! Everything is very quick to set up and get working.

I wanted to get a similar functionality to VSCode where dblclick on the handle expands and collapses the panels. This works, except that, because the click events are only being detected if I click right on the actual handle it becomes very annoying. Would it be possible / much work to make so that these events are fired if user clicks anywhere inside the hit area?

bvaughn:
Resolves #438


bvaughn:
Fix has been published in react-resizable-panels@2.1.8

---

PR #467, opened 2025-04-21
Feature: Add an option to propagate resize handle pointer events

OhadC:
Feature: Add an option to propagate resize handle pointer events

* New PanelResizeHandle prop: propagateEvents
* Add new attribute to support `propagateEvents`
* Attributes as const
* use `AbortController` for `updateListeners` (avoid the need to call `removeEventListener` on each element)
* Add & Update usage for all resize handle related attributes

Reason:
I need to add pointer events for an element that lives inside `PanelResizeHandle`
image

The problem is that the library listeners are catching the events and stopping them, and there is no other way to do that (because of the `capture` property introduced in https://github.com/bvaughn/react-resizable-panels/pull/338)

### DIFF
--- a/packages/react-resizable-panels/README.md
+++ b/packages/react-resizable-panels/README.md
@@ -44,7 +44,6 @@ import { Panel, PanelGroup, PanelResizeHandle } from "react-resizable-panels";
 - `setItem: (name: string, value: string) => void`
 
 `PanelGroup` components also expose an imperative API for manual resizing:
-
 | method                        | description                                                      |
 | :---------------------------- | :--------------------------------------------------------------- |
 | `getId(): string`             | Gets the panel group's ID.                                       |
@@ -73,7 +72,6 @@ import { Panel, PanelGroup, PanelResizeHandle } from "react-resizable-panels";
 <sup>1</sup>: If any `Panel` has an `onResize` callback, the `order` prop should be provided for all `Panel`s.
 
 `Panel` components also expose an imperative API for manual resizing:
-
 | method                   | description                                                                        |
 | :----------------------- | :--------------------------------------------------------------------------------- |
 | `collapse()`             | If panel is `collapsible`, collapse it fully.                                      |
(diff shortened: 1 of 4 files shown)


bvaughn:
I like a lot of the changes in this PR, but I don't love the added prop. I wonder if removing `stopPropagation` (and just calling `preventDefault`) would be sufficient to allow your use case _and_ address the original complaints from #338? Would need to actually revisit that PR and do a bit of testing.

bvaughn:
Thank you for confirming your use case.

@4nnikaf can you confirm the other?

bvaughn:
For now I'm going to move forward with the noncontroversial changes in this PR via #470

OhadC:
I didn't checked about that, but maybe we can check if the click was **inside** of the `intersectingHandles` (or only in their margins), and stop propagation only if the click is in the margin areas.
If this will work - is it fine by you? it will remove the need of the additional property.

bvaughn:
Let me consider it for a few minutes but I think that could work 🤔

bvaughn:
Just pushed an update that:
1. Only calls `stopPropagation` if the pointer event is outside of the resize handle as you suggested
2. Updates "pointermove" and "pointerup" event handlers to use capture and align with "pointdown" event handler. (I think this makes the most sense, though let me know if you disagree.)

OhadC:
Sounds great!
Thanks a lot!
Should I do the same `const **_ARRITBUTES = {}` for the other components?

bvaughn:
> Sounds great! Thanks a lot! Should I do the same `const **_ARRITBUTES = {}` for the other components?

I had been planning on doing this once we got this PR merged 😄

OhadC:
> > Sounds great! Thanks a lot! Should I do the same `const **_ARRITBUTES = {}` for the other components?
> 
> I had been planning on doing this once we got this PR merged 😄

I can do that
I saw another minor change that will make the code more readable :)

bvaughn:
I appreciate that, but I'm already doing it :)

bvaughn:
#471

bvaughn:
Published as part of 2.1.9

---

PR #470, opened 2025-04-27
Minor code cleanup: data attributes, `AbortController`

bvaughn:
Non-controversial portions of #467 (cc @OhadC)

* Attributes as const
* use `AbortController` for `updateListeners` (avoid the need to call `removeEventListener` on each element)
* Add & Update usage for all resize handle related attributes


bvaughn:
Published as part of 2.1.9

---

PR #473, opened 2025-04-27
Re-add insertRule() and fix conflict scenario

bvaughn:


bvaughn:
Published as part of 2.1.9

---

PR #474, opened 2025-04-27
Make the package ESM-only

Andarist:
I'm opening this PR just as a point of reference. It's up to you to decide at which point you'd like to abandon CJS. In the world with `require(esm)` (which is still pretty new) there is no strong reason to support it (unless you care about older versions of runtimes/tools/etc).

This might not play well with Jest (but given the ESM is being only introduced in your package here it might be fine). It's fixable but if you'd run into any issues with that I'd just recommend migrating to Vitest at this point. YMMV.

⚠️ this is a breaking change

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "@parcel/transformer-js": "^2.9.3",
     "@parcel/transformer-react-refresh-wrap": "^2.9.3",
     "@parcel/transformer-typescript-types": "^2.9.3",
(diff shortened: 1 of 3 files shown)


bvaughn:
I appreciate this. I think it's a reasonable thing to do.

bvaughn:
I merged into the wrong fork of `type-module` so let's move over to #478

PR #478, opened 2025-05-01
Make module ESM only; replace Jest with Vitest

bvaughn:
Builds on top of #474; special thanks to @Andarist for putting this branch together!


bvaughn:
Exciting milestone!

bvaughn:
Published as react-resizable-panels@3.0.0

packages/react-resizable-panels/package.json
```diff
@@ -71,19 +47,21 @@
...
+    "@vitest/ui": "^3.1.2",
     "eslint": "^8.37.0",
     "eslint-plugin-react-hooks": "^4.6.0",
-    "jest": "^29.7.0",
-    "jest-environment-jsdom": "^29.7.0",
+    "jsdom": "^26.1.0",
     "react": "experimental",
-    "react-dom": "experimental"
+    "react-dom": "experimental",
+    "save-dev": "0.0.1-security",
```
origami-z, line 63: Is this intentional...?

Andarist: No, it was accidentally introduced in https://github.com/bvaughn/react-resizable-panels/pull/476/files#diff-b9daed64243c63f6aac6e93e6d75abb73e84a246c863e14171ae0ba33775aaf7R63 but, luckily, this isn't a harmful package.

bvaughn: No. This is a result of me trying to maintain this stuff after hours when I'm half awake.

PR #479, opened 2025-05-02
Add worker-related conditions

Andarist:
This should improve support for Cloudflare Workers and Vercel Functions. Bundlers for workers often, annoyingly, prefer `browser` conditions over node/default - so when we include a browser condition, we, almost always, have to specify the worker conditions first in the context of React.


packages/react-resizable-panels/package.json
```diff
@@ -20,12 +20,18 @@
     ".": {
       "types": "./dist/react-resizable-panels.js",
       "development": {
+        "edge-light": "./dist/react-resizable-panels.development.edge-light.js",
+        "worker": "./dist/react-resizable-panels.development.edge-light.js",
+        "workerd": "./dist/react-resizable-panels.development.edge-light.js",
```
Andarist, line 25: Notice those are the same as the one for `node` - the bundle itself is reused. Those keys here are just listed in order of preference (remember those keys here are **ordered** - despite being just object keys)

bvaughn (approved):
"workerd" 😮‍💨 Whew, the JS toolchain can be exhausting sometimes

bvaughn:
Published in 3.0.1

Thanks!

Issue #483, opened 2025-05-08
[bug] workerd - Disallowed operation called within global scope


geemanjs:
Hey 👋 

Thanks for maintaining this library.

Probably just what you wanted after the most recent release - another workerd issue 👀. Albeit this one I think is a bit simpler. 

When running in a cloudflare worker / wrangler - I get the following:
```
Disallowed operation called within global scope. Asynchronous I/O (ex: fetch() or connect()), setting a timeout, and generating random values are not allowed within global scope
```

Having debugged the issue it spans from this line here where the AbortController is created in the global scope:
https://github.com/bvaughn/react-resizable-panels/blob/main/packages/react-resizable-panels/src/PanelResizeHandleRegistry.ts#L303

I think it will be a pretty simple fix (which is why I haven't shared a Minimal reproduction or more details about my setup). I'll take a look and see if I can get a PR together later to fix - but thought i'd raise in case anyone else stumbled upon it.

James

alekhinen:
Ran into this exact same issue. Looks like its a known problem over here too: https://github.com/cloudflare/workerd/issues/3657

geemanjs:
Nice find @alekhinen - feels like a workerd issue as there aren't any async operations occurring. Lets hope the fix comes soon 🙏

bjacobso:
I'm also running into this, to me this isn't a workerd issue, the library shouldn't really initialize an AbortController in the global scope.

bvaughn:
Fixed via #492 and just published via 3.0.3

PR #492, opened 2025-06-12
Fix compatibility with cloudflare workers

cloudkite:
cloudflare workers throw error if you initialize AbortController in global scope see: https://github.com/cloudflare/workerd/issues/3657, https://github.com/bvaughn/react-resizable-panels/issues/483


bvaughn (approved):
Oh! Nice find 🙇🏼 Thank you!

PR #528, opened 2025-11-16
Version 4

bvaughn:
To test a pre-release of version 4:
```sh
npm install react-resizable-panels@alpha
```

> [!NOTE]
> [Here](https://github.com/bvaughn/react-resizable-panels/blob/v4/CHANGELOG.md#migrating-from-version-3-to-4) is a version 3 to version 4 migration guide.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-resizable-panels",
-  "version": "4.0.0-alpha.5",
+  "version": "4.0.0",
   "type": "module",
   "author": "Brian Vaughn <brian.david.vaughn@gmail.com> (https://github.com/bvaughn/)",
   "contributors": [


isBatak:
Do you have any idea how to optimize nested layout re-renders?

In my opinion, only the root `Group` element should update when the root `Panel` is resized, since the sizing is now based on CSS variables. However, in the attached video you can see that all nested panels also re-render during the resize. That seems strange, because only the root group’s styles are being updated.

https://github.com/user-attachments/assets/2430aa17-1ae1-407b-a387-7b8d695ab12d

**EDIT:**
False alarm! 
It's because of this `onResize` handler on the `StyledPanel` https://github.com/bvaughn/react-resizable-panels/pull/528/files#diff-46f73d84bac2e080fe15faf8821126552550d877886ee01821751501a66389d2R27
When removed, everything runs smoothly 👌

https://github.com/user-attachments/assets/af400049-0132-41bf-846f-76dfcfaeeb8b

bvaughn:
> In my opinion, only the root Group element should update when the root Panel is resized, since the sizing is now based on CSS variables.

I think this might be a bit of premature optimization. `Panel` is a very light weight component. Shouldn't matter much whether  it renders or not, unless the content within a panel is expensive to re-render (but then again it could be memoized separately).

That being said, the re-render in this case is because my example `Panel` component indiscriminately attaches a `onResize` handler 😄 the library won't _observe_ the panel unless an external resize handler is passed in as a prop.

isBatak:
I was thinking it would be useful to have a `disabled` prop or something similar that completely stops resizing. This would make it easier to disable the resizing behavior on mobile screens and let users reorganize the same DOM elements to better fit mobile layouts. Is this something you’d consider adding?

bvaughn:
I have been considering that, and I haven’t totally decided against it, but
it feels a bit weird to render non-resizable resizable panels. Seems like
it would be better/more performant to just render divs.

bvaughn:
I should add that I think supporting a `disabled` prop on group seems fairly reasonable. I'm fairly against the idea of adding `disabled` props to resize handles or panels though, mostly because I don't feel the added complexity in the layout algorithm is worth it.

isBatak:
> I should add that I think supporting a `disabled` prop on group seems fairly reasonable. I'm fairly against the idea of adding `disabled` props to resize handles or panels though, mostly because I don't feel the added complexity in the layout algorithm is worth it.

Yeah this is exactly what I had in mind. Disabling it on the group level, no need for disabling panels or handles.

bvaughn:
> > I should add that I think supporting a disabled prop on group seems fairly reasonable. I'm fairly against the idea of adding disabled props to resize handles or panels though, mostly because I don't feel the added complexity in the layout algorithm is worth it.
>
> Yeah this is exactly what I had in mind. Disabling it on the group level, no need for disabling panels or handles.

Added.

lib/components/group/types.ts
```diff
@@ -0,0 +1,138 @@
...
+};
+
+export type DragState = {
+  state: "default" | "hover" | "dragging";
+  separatorId: string | undefined;
+};
+
+export type RegisteredGroup = {
+  defaultLayout: Layout | undefined;
+  direction: Direction;
```
isBatak: **suggestion:** rename `direction` to `orientation` because direction sounds similar to flex-direction, but uses different values, which could cause confusion.

bvaughn: I agree that "direction" is a bit overloaded with Flex, and the abbreviated "dir" style is also vaguely related. I'm not super worried about confusion though really, with the TypeScript types being bundled along, and the default (horizontal) layout being the most common by far. I think it's probably a non-issue.

I don't love the name "orientation" for this.

Another library of mine, `react-window`, switched from `direction: "horizontal" | "vertical"` to `layout` but only because I wanted to allow `direction: "ltr" | "rtl"` which... could be argued to apply here, but I think I'd rather not try to support that in this library.

isBatak: Will this convince you :D? 
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-orientation

bvaughn: That’s kind of compelling. Let me think on it.

lib/components/group/useDefaultLayout.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export function useDefaultLayout({
+  groupId,
+  storage = localStorage
+}: {
+  groupId: string;
+  storage?: Pick<Storage, "getItem" | "setItem">;
+}) {
+  const defaultLayoutRef = useRef<Layout | undefined | null>(null);
+  if (defaultLayoutRef.current === null) {
```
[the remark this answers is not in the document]
bvaughn: Ah, I'm assuming you _aren't_ using cookies (like the docs show), and are just using client-only `localStorage`?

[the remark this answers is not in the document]
bvaughn: I committed my Next and Vike integration projects (0e2f76b). Been meaning to do this anyway as I'll want to add some e2e tests and otherwise build out more integrations and examples for my own testing going forward.

isBatak:
Previously, panels had `onCollapse` and `onExpand` callbacks https://github.com/bvaughn/react-resizable-panels/blob/main/packages/react-resizable-panels/src/Panel.ts#L27.
What is the new recommended way to handle these callbacks?

isBatak:
Also, maybe I missed it, but did you completely drop the idea of PersistScript, or is it planned as a separate package?

bvaughn:
> What is the new recommended way to handle these callbacks?

`onResize` for the moment. The other two were just convenience wrappers.

With v4 you could just do:
```ts
onResize={(size) => {
  const isCollapsed = size === collapsedSize;
}}
```

Or:
```ts
onResize={() => {
  panelRef.isCollapsed();
}}
```

> did you completely drop the idea of PersistScript, or is it planned as a separate package?

I don’t know if it’s necessary to support that as well, at least not as part of this library. The more redundant ways of doing a thing the harder it is to maintain and test them all.

Thought that was the understood outcome of our chats on the other PR, eg https://github.com/bvaughn/react-resizable-panels/pull/521#issuecomment-3592942089

bvaughn:
FYI @isBatak I just published `react-resizable-panels@4.0.0-alpha.3` from commit f3e1580cd3990dc0eefc781a0de7f51d77d56d51 if you'd like to give it another spin.

PR #580, opened 2026-01-06
feat: add resize smoothing feature to Group component

arshaan-abh:
- Introduced `resizeSmoothing` prop to the Group component to allow for smooth layout changes using an exponential moving average (EMA).
- Implemented utility functions `normalizeResizeSmoothing` and `shouldNotifyLayoutChange` to handle smoothing logic and layout change notifications.
- Updated the Group component to utilize the new smoothing functionality, ensuring that layout changes are smoothed based on the provided prop.
- Enhanced the global state management to accommodate the new layout target and smoothing behavior.
- Added tests for the new smoothing functionality and its associated utilities.
- Updated documentation and examples to demonstrate the new `resizeSmoothing` feature in action.


arshaan-abh:
I had to tweak `compile-docs.ts` and `compile-examples.ts` to make the compile scripts work on Windows (path normalization issues). If you’d prefer, I can revert those changes and keep the workaround local. Just let me know.

arshaan-abh:
https://github.com/user-attachments/assets/a425c872-28ea-4612-a658-1086c23275cc

bvaughn:
Thanks for sharing this PR and the idea! I'm not completely sure I understand the problem it's solving though. Can you spell it out a bit more clearly (maybe including a before and after video)? The resize behavior already seems "smooth" to me as it just follows the user pointer.

> I had to tweak `compile-docs.ts` and `compile-examples.ts` to make the compile scripts work on Windows (path normalization issues). If you’d prefer, I can revert those changes and keep the workaround local. Just let me know.

Would you be willing to post a PR to [react-lib-tools](https://github.com/bvaughn/react-lib-tools)? Would be nice if this package worked for developers on Windows too.

arshaan-abh:
> Can you spell it out a bit more clearly (maybe including a before and after video)?

Sure! Here’s a video where I toggle the resizeSmoothing prop to see the difference:

https://github.com/user-attachments/assets/fa3303f9-1115-4f70-8390-78d2fac27726

The `resizeSmoothing` controls how layout changes are eased using an exponential moving average, smoothing resize transitions across all interactions. You’ll notice the effect most when panels are collapsible or when you set a lower resizeSmoothing value for extra‑smooth motion.

arshaan-abh:
> Would you be willing to post a PR to [react-lib-tools](https://github.com/bvaughn/react-lib-tools)? Would be nice if this package worked for developers on Windows too.

I’d be happy to, though I won’t be able to get to it right away.

arshaan-abh:
Here's another example with `resizeSmoothing` set to `0.025`:

https://github.com/user-attachments/assets/8445fed7-188f-4ed4-afcf-fd323a899f0e

bvaughn:
Interesting! Thanks for the video.

I can't really spot the difference in your first video. (It might be something that's hard to _see_ unless you're the one moving the pointer.) The second video is _interesting_ but... also could seem laggy/unresponsive 🤔  It's not clearly _better_ just different.

On the one hand, it's optional (opt-in) but on the other hand every new prop and mode this library supports, the more edge cases it needs to handle and the harder it gets to maintain. I'm not sure how I feel about this feature idea.

> > Would you be willing to post a PR to [react-lib-tools](https://github.com/bvaughn/react-lib-tools)? Would be nice if this package worked for developers on Windows too.
> 
> I’d be happy to, though I won’t be able to get to it right away.

Totally! 😄

bvaughn:
I checked out the branch and I can definitely notice the difference when I'm the one moving the mouse. It feels _less responsive_ to me though.

bvaughn:
I like it better when I'm using a collapse button (or typing Enter on the keyboard while focused on a separator) though that has the downside of arguably being broken while the animation is in progress (typing Enter again does nothing until the animation completes, when I think it _should_ reverse the in-progress collapse)

arshaan-abh:
The original idea came from the shadcn sidebars. I paired them with the shadcn resizable component (which uses this library) so the sidebar could be both resizable and collapsible. That worked really well and felt like a nice improvement for shadcn sidebars overall.

One thing that stood out, though, was that when the sidebar collapses, it happens instantly. There’s no transition or animation. That felt a bit abrupt to me. I initially wanted to add a transition just for the collapse action in this library, but having only the collapse animated felt inconsistent with the drag-and-drop interaction, and that also felt a little off (though I still think it’s an idea worth revisiting).

So instead, I focused on adding smoothness to the entire interaction.

> typing Enter again does nothing until the animation completes, when I think it should reverse the in-progress collapse

I agree! I can fix that.

bvaughn:
I understand.

After thinking about it more last night, I don't think this is a feature I want to add to the library.

arshaan-abh:
I understand. Thanks for your time.

bvaughn:
Absolutely. Thank you for yours as well.

Would still be happy to review any changes needed to make the supporting library work on windows if you feel like submitting a PR there

Issue #574, opened 2026-01-02
Cannot click on selected text to move cursor when Group is rendered


JinJieBeWater:
## Description

When a `Group` component is rendered on the page, clicking on selected text to move the cursor doesn't work. This affects text anywhere on the page, not just inside the Group.

## Steps to Reproduce

1. Go to https://react-resizable-panels.vercel.app/
2. Select some text on the page (either in the demo panels or outside)
3. Try clicking within the selected text to move the cursor
4. The cursor will not move to the clicked position

## Expected Behavior

Clicking on selected text should move the cursor to that position, just like normal text behavior.

## Actual Behavior

Clicking on selected text has no effect - the cursor position doesn't change.

bvaughn:
Am I missing something? Seems to work fine for me.

https://github.com/user-attachments/assets/07b2282c-bd08-47ff-a57f-837f5ecfab35

https://github.com/user-attachments/assets/a7e019ea-8902-4d55-9197-244ab7c2e0ab

Try restarting your browser. Maybe you're seeing a browser bug of some sort.

JinJieBeWater:
It seems that our mouse behavior differs.

When using **Google Chrome**, the issue consistently occurs.
I tested the same interaction in **Zen Browser (Firefox-based engine)**, and the behavior is normal there.

To rule out profile- or extension-related factors, I also tested Chrome with:

* A **new browser profile**
* **All extensions removed**
* **Default settings restored**

The issue still persists in Chrome.

Chrome
https://github.com/user-attachments/assets/484b87cd-d8bd-4c54-b926-8585d17a609d

Zen 
https://github.com/user-attachments/assets/2feba856-beb0-4e1a-b3bb-f15a7083bb6c

This is a bit confusing.

---

**Environment**

* Operating System: **Windows 26200.7462**
* Browsers tested:

  * **Google Chrome** (issue reproducible)
  * **Zen Browser (Firefox-based engine)** (works as expected)

bvaughn:
The video I recorded above was made using Chrome, although I'm using a Mac.

JinJieBeWater:
I tested this on my friend’s computer as well. Using Chrome on Windows shows the same issue, so it appears to be a Chrome-on-Windows–specific problem.

bvaughn:
Interesting!

Can you repro it on a website that isn't my documentation site? (I'm trying to rule out a problem with the package itself vs just the website I made to document it.)

For example, does it happen here?
[...]

PR #595, opened 2026-01-09
Only call preventDefault on "pointerup" events that were handled by this lib

### LINKED ISSUE #594: v4.3.1 capture phase pointer events breaks Radix UI Select inside ResizablePanel
Bug Report:

  ---
  Description

  After upgrading from 4.2.0 to 4.3.1+, Radix UI Select components inside ResizablePanel no longer respond to clicks. The dropdown opens, options are visible and can be focused via keyboard, but clicking on an option does nothing - the selection doesn't change and the dropdown doesn't close.

  This regression was introduced in v4.3.1 by PR #589 which switched to capture phase for pointerdown and pointerup events.

  Steps to Reproduce

  1. Place a Radix UI Select component inside a ResizablePanel
  2. Click to open the Select dropdown
  3. Try to click on any option

  Expected Behavior

  Clicking an option should select it and close the dropdown.

  Actual Behavior

  Click events on Select options are not registered. The option receives focus/hover state but the click handler never fires.

  Environment

  - react-resizable-panels: 4.3.1 / 4.3.2 (broken)
  - react-resizable-panels: 4.2.0 (works)
  - radix-ui: latest
  - React: 19

  Workaround

  Downgrade to v4.2.0

  Analysis

  The capture phase event listeners added in PR #589 intercept pointer events before they can reach Radix UI's Select options. Radix UI portals Select content to the document body, but the capture phase listeners on the ResizablePanelGroup catch the events first. 

_Note: 4.3.0. works fine too, the regression happens with 4.3.1_

bvaughn:
Resolves #594

### DIFF
--- a/lib/global/event-handlers/onDocumentPointerUp.ts
+++ b/lib/global/event-handlers/onDocumentPointerUp.ts
@@ -8,8 +8,6 @@ export function onDocumentPointerUp(event: PointerEvent) {
     return;
   }
 
-  event.preventDefault();
-
   const { interactionState } = read();
 
   switch (interactionState.state) {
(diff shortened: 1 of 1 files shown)


JinJieBeWater:
Oh, that also solves #574

Issue #368, opened 2024-06-27
Panels not collapsing completely at window edges


philippklodt:
Repro at: https://codesandbox.io/p/sandbox/n9vs2g

https://github.com/user-attachments/assets/dcd6548a-0323-43a5-965d-b5046abb5ace

> [!NOTE]
> I was not able to reproduce this locally, even with an iframe, until I set the `sandbox` attribute.

<details>
  <summary>Original description</summary>
  
When trying to collapse a panel completely by dragging beyond the viewport/window, panels are sometimes not collapsed completely.

This happens when the panel-group's edge corresponds to the viewport edge (i.e. no margin).

I adapted one of the codesandbox samples and created a reproducible demo here:
https://codesandbox.io/p/sandbox/react-resizable-panels-forked-r78ckq?file=%2Fsrc%2FApp.js%3A18%2C19-18%2C41

Adding some margin/padding around the panel-group can be used as a workaround, but this is sometimes undesired -- especially in my case where I was trying to implement resizable app-sidebars.
</details>

bvaughn:
Makes sense.

Open to reviewing a PR if you'd like to suggest a fix!

philippklodt:
Thanks for the quick reply, would love to help, but currently I can't say when I'll have time to look into it.

bvaughn:
I'm not able to repro the behavior described in the Code Sandbox you linked to. Any chance you could share [a Replay](https://replay.io/) of the bug?

bvaughn:
Actually I think the reason I can't repro this is that I (hopefully) fixed it in [2.0.21](https://www.npmjs.com/package/react-resizable-panels/v/2.0.21). If it's still broken for you on the latest release, let me know. I think it might be good now though!

---

❤️ → ☕ [givebrian.coffee](http://givebrian.coffee/)

philippklodt:
Thanks for looking into it again.
I updated the version in the codesandbox to `2.0.22` and can still observe the issue.
Unfortunately I wasn't able to get replay.io to work because of some auth issue.

I made a video which may help at least to better explain the steps to reproduce:
https://github.com/user-attachments/assets/c7757869-0a8a-4d93-9c44-815ad5116db5

As you can see in the video, when I start dragging over the resizer-element, everything works fine.
If I start dragging outside the resizer-element (to its right), and the panel-group as a whole is very close to the right viewport edge, the issue persists.
I guess it's because outside the viewport there are of course no mouse events.
[...]

PR #609, opened 2026-01-22
fix #608: bidirectional cursor flashes while dragging

### LINKED ISSUE #608: Bidirectional move cursor flashes while dragging
I appreciated the attention to detail of having a bidirectional move cursor when using nested panels. :)

But in practice, during the move, the cursor flashes back and forth:

Image

It's actually even more jittery in real life than what the gif managed to capture.

It's reproducible on the documentation page, even with panels that are basically empty: https://react-resizable-panels.vercel.app/examples/nested-groups The behaviour feels a little better there, I think because my layout is more complex and takes slightly longer for the browser to recalc everything. 

I tested in Chrome 144.0 and Firefox 146.0

When dragging the same panels in only a single direction (i.e. not the intersection point) it works fine.

There must be a way to lock the cursor.

michaelboyles:
Fixes #608.

It may not be correct to remove this call, but it doesn't seem to do anything except make the cursor jittery.

Before:

before

After

after

(Ignore colours flashing a bit on the 2nd one - it's just an artifact of the gif)

### DIFF
--- a/lib/global/event-handlers/onDocumentPointerLeave.ts
+++ b/lib/global/event-handlers/onDocumentPointerLeave.ts
@@ -7,7 +7,6 @@ export function onDocumentPointerLeave(event: PointerEvent) {
   switch (interactionState.state) {
     case "active": {
       updateActiveHitRegions({
-        document: event.currentTarget as Document,
         event,
         hitRegions: interactionState.hitRegions,
         initialLayoutMap: interactionState.initialLayoutMap,
--- a/lib/global/event-handlers/onDocumentPointerMove.ts
+++ b/lib/global/event-handlers/onDocumentPointerMove.ts
@@ -37,7 +37,6 @@ export function onDocumentPointerMove(event: PointerEvent) {
       }
 
       updateActiveHitRegions({
-        document: event.currentTarget as Document,
         event,
         hitRegions: interactionState.hitRegions,
         initialLayoutMap: interactionState.initialLayoutMap,
--- a/lib/global/utils/updateActiveHitRegion.ts
+++ b/lib/global/utils/updateActiveHitRegion.ts
@@ -6,21 +6,18 @@ import {
   CURSOR_FLAG_VERTICAL_MIN
 } from "../../constants";
 import type { Point } from "../../types";
-import { updateCursorStyle } from "../cursor/updateCursorStyle";
 import type { HitRegion } from "../dom/calculateHitRegions";
 import { update, type MountedGroupMap } from "../mutableState";
 import { adjustLayoutByDelta } from "./adjustLayoutByDelta";
 import { layoutsEqual } from "./layoutsEqual";
 
 export function updateActiveHitRegions({
-  document,
   event,
   hitRegions,
   initialLayoutMap,
   mountedGroups,
   pointerDownAtPoint
 }: {
-  document: Document;
   event: {
     clientX: number;
     clientY: number;
@@ -119,6 +116,4 @@ export function updateActiveHitRegions({
     cursorFlags,
     mountedGroups: nextMountedGroups
   });
-
-  updateCursorStyle(document);
 }


bvaughn:
The code in question was added as part of the fix for #368 🤔

There are a few e2e tests that cover cursor behavior, and some were broken by this change. The [contributing guide](https://github.com/bvaughn/react-resizable-panels/blob/main/CONTRIBUTING.md#running-tests-locally) explains how to run them.

bvaughn:
Removing this line in particular would prevent the cursor from updating while a resize is in progress, which would be a significant UX regression I think.

https://github.com/user-attachments/assets/9e3f7ca7-7be1-475a-a356-4ecff7093fce

The cursor should react and update, like this:

https://github.com/user-attachments/assets/ac03d271-fcee-4707-9983-9d2f957c1e09

michaelboyles:
Sorry for not running the e2e tests. I did make sure to run the tests before submitting, but it must have only been the unit tests.

> Removing this line in particular would prevent the cursor from updating while a resize is in progress

It wasn't clear to me precisely what the cursor behaviour is supposed to be. Even ignoring the jitter, I don't really get it.

I will try to infer based on what I saw below. Correct me if I'm wrong!

Take this example, where I grab the same point as before and drag it way above the panels, so that 'top' can't get any shorter, and likewise the inner 'left+right' can't get any taller.

ezgif-3a186942ca21afcf

The cursor switches between a few states, but the one it settles on for perhaps the longest is north-south. I don't know why east-west is excluded because the left and center sections are never at their smallest in that clip. Lateral movement at that point will always do something, but the cursor is telling me the opposite.

It seems like what you're trying to do is update the cursor so at any point in time, it tells you which ways you can move the mouse that would cause the panel sizes to change. Is that the intention? If so, in my example, the cursor should show East-West-South from the point the cursor moves above the panels. There isn't a cursor which excludes only North, so that would not be possible.

What you call a UX regression, I actually feel might be an improvement. Keeping the same cursor you had when you started dragging feels simple and intuitive to me.

bvaughn:
> It seems like what you're trying to do is update the cursor so at any point in time, it tells you which ways you can move the mouse that would cause the panel sizes to change. Is that the intention?

Yes. This is how UIs like VS Code work and I think it's helpful for the cursor to reflect when the pointer has moved too far past the bounds of the group to resize it anymore.

> What you call a UX regression, I actually feel might be an improvement. Keeping the same cursor you had when you started dragging feels simple and intuitive to me.

We disagree here :)

michaelboyles:
> This is how UIs like VS Code work

I tested the same kind of behaviour in VS Code (admittedly a kind of old version - I don't use it), where I drag the intersection point above the top point of the panels. The cursor was still the 4-way one. Is it different for you? I didn't observe the cursor change from the initial one wherever I dragged it.

vscode-loop

Sorry if it's too zoomed in to get the full idea - github complained about the size otherwise. The blue-ish area is a ribbon that runs above any of the panels I'm moving, as you can see by the light blue panel indicator that goes under it.

> We disagree here

That's fine, but can you justify why your lib shows North-South when it can move side to side? That just seems wrong. I'm not trying to get distracted onto a different issue (if that is an issue), but it's relevant to me understanding what the behaviour is supposed to be.

bvaughn:
> That's fine, but can you justify why your lib shows North-South when it can move side to side?

There _is_ a bug in the cursor updates, as is being discussed over on #608 (see my latest comments there)

The _intended behavior_ of the cursor is what's asserted on in the e2e tests. It's also the behavior I see in VS Code:

https://github.com/user-attachments/assets/c1947257-b472-48f3-b20c-35fdec79952a

This is the behavior that I think provides the best UX, so that's what the library implements.

PS you can always set `disableCursor` on the Group and do whatever you want with the cursor, if you'd prefer something else.

michaelboyles:
That video only shows VS Code's lateral movement. What does the cursor do when you have panels that move in all 4 directions? I showed it already, but maybe it's different for your OS/version.

I'm agreed with you now that there are potentially valid updates while the cursor is moving, which makes the change in this PR wrong. :)

What I think you're missing, and the behaviour that VS Code seems to have for me, is that it only applies when the movement is on a single axis.

You _can_ have a cursor that starts east-west, then changes to east-only when you've moved all the way left. Because [all the 1-direction and 2-direction cursors are available](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/cursor)!

You _can't_ have a cursor that starts in 4-directions then omits directions based on where the cursor is, because there are no 3-direction cursors that exist (e.g. South-East-West as would be needed previously). So it seems that when movement is on both axes, VS Code does not try to update it.

bvaughn:
I don't really want to argue what the "correct" UX on this PR. That doesn't seem productive. I've implemented what I think is the best UX. If you disagree, you can disable it (like I mentioned in my previous comment) but I don't plan to change it.

I _would_ like the fix the bug, and I'm trying to get to the bottom of it (as the comment trail on the original bug report indicates) but this is a bit of a distraction.

michaelboyles:
Ok that's fine. I'm not trying to argue what the UX ought to be, but you chose VS Code as a reference and I was just noting that the behaviour doesn't match for me (in the 4-direction case). If you're expecting that, that's fine.

Would you be annoyed if I opened that as a separate issue or shall I just drop it? You'd be welcome to close it, of course. I'll sleep on it either way. Perhaps I'll wait and see what it's like after any fixes for the jittering too.

bvaughn:
Sure, that's fair. I mentioned VS Code as an example of a high-profile app that updates the cursor while you move. I was responding specifically to your comment above:

> Keeping the same cursor you had when you started dragging feels simple and intuitive to me.

> Would you be annoyed if I opened that as a separate issue or shall I just drop it?

I don't plan on changing the intersecting panels cursor UX for this library. I think the way it works is intuitive and helpful.

https://github.com/user-attachments/assets/536e573b-4bf5-46e7-a477-b9a0fd13d019

If I had received a lot of feedback that it was confusing to people then I would reconsider, but this is the only time I've heard that.

michaelboyles:
My last comment, promise.

If I'd seen the behaviour you showed there, I doubt I would have thought twice about it, so maybe I'm overthinking it now I'm in the weeds. 

Just for "fun" I tried to do the same movement as you.

https://github.com/user-attachments/assets/30c92e60-887f-4cd5-a0de-0b17bce66840

It seems for whatever reason that it's much more "random" for me. If you find a fix and want me to verify, feel free to reach out. Happy to pull a branch or whatever 

Thanks for looking into it

bvaughn:
> It seems for whatever reason that it's much more "random" for me.

Yes, that is the bug. There _is_ a bug. I've mentioned that above 😄 

Let's move this conversation over to the original issue? I've left several comments there about the bug. I think I have a fix for it locally.

PR #632, opened 2026-01-27
Add data attribute to disable hover & drag on elements near separator (#630)

### LINKED ISSUE #630: Support buttons in separators
Thanks for a great library!

What's the recommended approach to put buttons in separators? What I want is the same as I get in Chrome Devtools, where you can resize the panel anywhere on an empty space of the toolbar.

Like this:

https://github.com/user-attachments/assets/75ed1b12-0e5f-4d43-ad73-cc987464cfbb

I've tried to implement this myself, but there are a couple of issues that are annoying/difficult to work around.

Right now this css is inserted on hover:

```css
*, :hover {
  cursor: ns-resize !important;
}
```

This is possible to override with custom classes on all elements in the separator. But this is very cumbersome to do if using libraries such as tailwind where global styles take precedence over layers. 

And if you override the hover style on the buttons, the cursor gets an incorrect style when you resize and move the cursor over them. It should keep `cursor: ns-resize` while resizing.

Happy to provide a codesandbox if needed!

k-ode:
Added a check that ignores elements with the data attribute `data-resize-disabled`. This is for disabling resizing when interacting with something like a button near or over a separator.

I tried making the least disruptive change I could think of, but maybe this prop should actually be exposed by the library with a hook or something. Might not be needed though if it's more of an edge case.

Fixes #630

### DIFF
--- a/lib/global/utils/isViableHitTarget.ts
+++ b/lib/global/utils/isViableHitTarget.ts
@@ -20,8 +20,15 @@ export function isViableHitTarget({
   hitRegion: DOMRect;
   pointerEventTarget: EventTarget | null;
 }) {
+  if (!isHTMLElement(pointerEventTarget)) {
+    return true;
+  }
+
+  if (pointerEventTarget.closest("[data-resize-disabled]")) {
+    return false;
+  }
+
   if (
-    !isHTMLElement(pointerEventTarget) ||
     pointerEventTarget.contains(groupElement) ||
     groupElement.contains(pointerEventTarget)
   ) {


k-ode:
Just realized something: This should only block on hover and mousedown, not when we are already resizing.

Going to try and fix this later.

bvaughn:
Thanks for sharing this idea!

I think...this feels like a bit too much of an edge case for me to try adding custom data-attribute checks to support, _but_ I left a comment (https://github.com/bvaughn/react-resizable-panels/issues/630#issuecomment-3807834446) on the original issue showing how you can implement this with the current version of the library.

PR #636, opened 2026-01-28
build: add sonarjs lint plugin, fix lint errors

michaelboyles:
I noticed in a recent commit that you [removed some unused collections](https://github.com/bvaughn/react-resizable-panels/commit/5a11ea6562cbbea7ff5a42885d64e7969a1f38c4#diff-53ce94c55647fe0f3addc83cf2298ca14ddac4b13403434adf6b6fa6cc9a4afbL15-L18) - they were only added to, never accessed.

There was another instance of this, so I added sonarjs to eslint which has a nice check for unused collections.

I would personally enable `sonarjs/no-small-switch`. There are instances of switches with only one case, but I know this is some people's preference to keep a clean history, especially if they anticipate additional cases being added later.

Likewise I'd enable `sonarjs/no-nested-conditional` but I left it off in case you prefer this style.

`@types/sharp` was removed for an unrelated reason:

```
WARN  deprecated @types/sharp@0.32.0: This is a stub types definition. sharp provides its own type definitions,
so you do not need this installed.
```


michaelboyles:
Sure. I can take or leave that change too. Ends up actually being more lines, and it's more hassle if you want to put a breakpoint on only one case.

But you can see why it made the suggestion: both the branch bodies are the same.

Do you prefer I turn that check off or keep it on, but suppress the error for those 2 cases?

bvaughn:
I'd prefer that check be disabled.

michaelboyles:
Sure, all done

bvaughn (approved):

bvaughn:
Rebasing #598 and I'm noticing a couple of surprisingly opinionated rules in Sonar, like `sonarjs/function-return-type`, e.g.
Screenshot 2026-01-28 at 6 16 11 PM

A function that returns a possible undefined/null value is very common and reasonable IMO.

michaelboyles:
Ah, sorry if it was annoying. I haven't set it up it since my prev job so I couldn't remember how much or what was turned off.

Agree on that warning. It's a shame because there's plenty of good checks, but the preset is pretty bad

bvaughn:
Oh no worries at all! I was just sharing to share 😆 I've turned down a few more of their default rules and it's more manageable now

PR #649, opened 2026-02-01
Optimization: Replace `useForceUpdate` with `useSyncExternalStore`

bvaughn:
This should work around an undesirable edge case with "click" events that was reported in #627, while still maintaining the performance optimization added in #638.

I have published this change as alpha version `react-resizable-panels@4.5.9-alpha.0`

@michaelboyles Please verify this fixes the "click" issue you reported

@evanfrawley Sorry to ask, but can you confirm this change doesn't negatively impact the issue you reported in #637?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- [649](https://github.com/bvaughn/react-resizable-panels/pull/649): Optimization: Replace `useForceUpdate` with `useSyncExternalStore`
+
 ## 4.5.8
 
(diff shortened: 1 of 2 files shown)


evanfrawley:
I will be able to test in a few hours. Thanks for the heads up!

bvaughn:
Much appreciated!

evan-sift:
Its looking great 👍

bvaughn:
Thanks, Evan!

michaelboyles:
Yep, it solves the click issue too.

bvaughn:
Thanks, Michael

PR #703, opened 2026-04-06
fix: edge case collapse last panel

hugs7:
When collapsing the a panel within a group where all other (previous) panels are already collapsed, there's currently a bug where the delta calculation cascades to cause the first panel to expand. This is undesired since the first panel should remain collapsed given we are not touching it.

This fix handles the edge case by computing the remaining space percentage for the last panel to take up, thus leaving previous panels undisturbed.

By "previous" panels here I mean panels above (vertical mode) or to the left (horizontal) mode.

### DIFF
--- a/lib/global/utils/getImperativePanelMethods.ts
+++ b/lib/global/utils/getImperativePanelMethods.ts
@@ -116,10 +116,7 @@ export function getImperativePanelMethods({
           .slice(0, index)
           .reduce((total, panel) => total + prevLayout[panel.id], 0);
 
-        const fallbackLayout: Record<string, number> = {};
-        for (const key of Object.keys(prevLayout)) {
-          fallbackLayout[key] = prevLayout[key];
-        }
+        const fallbackLayout = { ...prevLayout };
         fallbackLayout[panelId] = formatLayoutNumber(100 - occupiedByPrevious);
 
         const nextLayout = validatePanelGroupLayout({
--- a/lib/global/utils/getImperativePanelMethods.ts
+++ b/lib/global/utils/getImperativePanelMethods.ts
@@ -1,4 +1,8 @@
-import type { PanelImperativeHandle } from "../../components/panel/types";
+import type { Layout } from "../../components/group/types";
+import type {
+  PanelConstraints,
+  PanelImperativeHandle
+} from "../../components/panel/types";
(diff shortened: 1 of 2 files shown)


hugs7:
Hi @bvaughn , could you kindly have a look at my PR?

lib/global/utils/getImperativePanelMethods.ts
```diff
@@ -71,6 +76,64 @@ export function getImperativePanelMethods({
...
+    const prevSize = getPanelSize();
+
+    const index = panels.findIndex((current) => current.id === panelId);
+    const isFirstPanel = index === 0;
+    const isLastPanel = index === panels.length - 1;
+
+    const allPreviousCollapsed =
+      isLastPanel &&
+      nextSize < prevSize &&
+      (isFirstPanel ||
```
hugs7, line 108: I also noticed this assertion error can happen when collapsing the one and only panel in a group, since the `delta` value otherwise goes to `100` causing it to index at `-1` [here at adjustLayoutByDelta on line 142](https://github.com/bvaughn/react-resizable-panels/blob/main/lib/global/utils/adjustLayoutByDelta.ts#L142)

Image

hugs7: I've addressed this in this PR via the following logic: If there's only one panel in the group, we can still collapse it but leave it's flex-grow value at 100%

hugs7:
Unit tests

image

bvaughn:
I appreciate this contribution! Thanks!

Just wanted to let you know that I am moving Apartments today and tomorrow
so it will likely be a bit before I can review these changes. I’ll try to
get to them as soon as I can, though.

hugs7:
No rush. Please take your time to review

bvaughn:
Have not reviewed the code yet. TBH I'm still probably too exhausted to do this, BUT– I would like to see at least two new unit tests to be included w/ this sort of PR (for the edge cases you mention)

If you're willing to add them, I'd appreciate it!

hugs7:
Sure can do - please allow me some time

hugs7:
HI @bvaughn , apologies for delay - I've added some unit tests

bvaughn:
Thank you!

bvaughn:
This change has been release as part of 4.11.0.

Issue #714, opened 2026-05-05
Resize broken in 4.8+ on Vite dev / React Router SSR (panels keep flex-basis from defaultSize)


olegp:
I'm hitting a regression on resize where panels visibly stop resizing while the
library's internal state still updates. Reproduces on 4.10.0 and 4.11.0.
Pinning to 4.7.6 makes it work again. I haven't bisected 4.8.0/4.9.0
individually.

## Setup

- Vite 7 dev server with React Router 7 SSR (`@react-router/dev`)
- React 19.2
- Three panels (`<Panel defaultSize="15%" />`, `<Panel defaultSize="55%" />`,
  `<Panel defaultSize="30%" />`) with two `<Separator>`s between them.
- `useDefaultLayout({ id: "tables", storage: localStorage })` on the `<Group>`.

## Expected

After hydration, each Panel renders with `flex: <size> 1 0px`. Dragging a
separator updates these flex values and panels visibly resize.

## Actual

After SSR (introduced in #696), each Panel server-renders with
`flex-basis: <defaultSize>`. After hydration in our setup the Panel never
re-renders to drop `flex-basis` and pick up `flex-grow: <size>`. Dragging a
separator does update the library's state (`aria-valuenow` changes,
`flex-grow: <new size>` is written on the panel), but `flex-basis: 15%` /
`55%` / `30%` stays. Since the basis values sum to 100%, the browser honors
basis and ignores grow, so panels look frozen.

## DOM snapshot

Working (production build, 4.7.6):

```
style="...; display: flex; flex: 15 1 0px; overflow: visible;"
```

Broken (Vite dev, 4.10.0):

```
style="...display:flex;flex-basis:15%;overflow:visible"
```

After dragging in 4.10:

```
style="...display:flex;flex-basis:15%;overflow:visible;flex-grow:30.828"
```

(Notice both `flex-basis: 15%` and `flex-grow: 30.828` are set, which gives
no visible resize.)

The same code in a production build (Vite + Rollup) on 4.7.6 doesn't exhibit
the issue: the post-hydration render correctly drops `flex-basis` and the
panels show `flex: <size> 1 0px`. So this may be specific to dev-mode SSR /
hydration timing rather than a pure rendering bug.

## Version range

- 4.7.6: works
- 4.10.0, 4.11.0: confirmed broken in our setup
- 4.7.5, 4.8.0, 4.9.0: not tested but presumed affected since #696 shipped in
  4.7.5

[...]

PR #715, opened 2026-05-11
Fixed Panel condition to properly handle explicit defaultSize

aidenfoxx:
When setting `defaultSize` explicitly to 0 (hidden), the layout for other panels breaks. This is due to the following code:
```typescript
// When `defaultSize` is explicitly 0, this falls through to the `flexGrow` condition, causing
// the initial render calculations to be incorrect. This can be solved by explicitly checking
// `defaultSize !== undefined`
} else if (defaultSize) {
    panelStyles = {
      flexGrow: undefined,
      flexShrink: undefined,
      flexBasis: defaultSize
    };
} else {
    panelStyles = { flexGrow: 1 };
}
```

Example case: https://codesandbox.io/p/sandbox/99tvsj

### DIFF
--- a/integrations/next/app/page.tsx
+++ b/integrations/next/app/page.tsx
@@ -65,11 +65,12 @@ export default async function Home() {
       <DefaultSize defaultSize="100px" />
       <DefaultSize defaultSize="25vw" />
       <DefaultSize defaultSize="15rem" />
+      <DefaultSize defaultSize={0} />
     </div>
   );
 }
 
-function DefaultSize({ defaultSize }: { defaultSize: string }) {
+function DefaultSize({ defaultSize }: { defaultSize: number | string }) {
   return (
     <Group className="h-25 gap-2">
       <Panel className="bg-slate-800 rounded rounded-md p-2" id="left">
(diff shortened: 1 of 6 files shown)


bvaughn:
Is this the same as the issue being discussed in #714?

aidenfoxx:
@bvaughn I don't believe so, and re-reading my PR description I did a bad job articulating the problem. I re-wrore the description, and updated the codesandbox example to include a way to trigger and untrigger this issue. Mine is extremely minor and with a simple workaround (setting `defaultSize={1}`).

bvaughn:
Conceptually, swapping a falsey conditional with an explicit `!== undefined` check makes sense, but I'm still a bit confused about this bug and why it doesn't show up in one of my e2e tests, like this one:

https://github.com/bvaughn/react-resizable-panels/blob/85714d9811b7f984024a4c84a1dc81c7ad6ec9ab/integrations/tests/tests/default-sizes.spec.tsx#L25-L41

The layout rendered by that test is 200px, 0px, and the remaining width as I would expect.

Your Code Sandbox also looks a little odd in that the order of elements in the outer group is: panel, panel, separator.

That being said, looking at the rendered output actually looks expected to me. The panel renders with a default size of 0 or 1, whichever value you pass.

Can you include a video or some more context about what's not working correctly? I want to make sure I fully understand the problem and that we can add a unit tests for whatever this is.

bvaughn:
Ah, I see it. Updating my [layout shift e2e test](https://github.com/bvaughn/react-resizable-panels/blob/main/integrations/tests/tests/default-layout.spec.tsx) to render a panel with `defaultSize={0}` shows the shift.

bvaughn:
Published in react-resizable-panels@4.11.1

Thanks!

PR #721, opened 2026-06-12
feat(Group): pass isUserInteraction flag to onLayoutChanged (closes #716)

### LINKED ISSUE #716: Pass event source to `onLayoutChanged` so consumers can distinguish user drag from library-driven layout changes
## Description

`onLayoutChanged` fires for any layout change with no way for the consumer to tell **why** it fired:

1. User dragging a separator
2. `minSize` / `maxSize` clamp triggered by a container width change
3. Parent re-render that produced new `defaultSize` props
4. Mount-time echo
5. Programmatic `group.setLayout(...)`

When the callback is used to persist layout to a **shared** backend (multi-user CMS / page configuration), a clamp on one user's narrow viewport gets written back as the canonical layout. Every other user — even on a wide viewport where the clamp would not apply — inherits the clamped flex values. A per-user, viewport-specific constraint pollutes the shared record.

This is reproducible reliably when `minSize` / `maxSize` are configured in `px` units. Percentage constraints scale with the container so they remain satisfied across width changes; pixel constraints are absolute, so any container shrink below the sum of pixel `minSize`s forces a clamp that is reported through `onLayoutChanged` indistinguishably from a real drag.

## Feature Request / Bug Report

Expose the cause of the fire on the `onLayoutChanged` callback:

```ts
type LayoutChangeCause =
  | 'user-drag'            // pointerup after a separator drag
  | 'user-keyboard'        // keyboard resize, if supported
  | 'constraint-clamp'     // min/max size clamp due to container resize
  | 'default-size-change'  // parent re-rendered with new defaultSize props
  | 'mount'                // initial layout computation
  | 'programmatic'         // group.setLayout() called from user code

onLayoutChanged?: (layout: Layout, meta: { cause: LayoutChangeCause }) => void
```

Backwards-compatible: existing handlers ignoring the second argument keep working.

## Steps to Reproduce

1. Configure a `Panel` with a pixel `minSize` (e.g. `minSize="200px"`).
2. Wire `onLayoutChanged` to persist the resulting layout to a shared backend.
3. Resize the surrounding container below the total pixel `minSize` — toggle a sidebar, resize the browser, mount a wide admin panel, etc.
4. The library clamps the layout to satisfy the constraint and fires `onLayoutChanged` with the clamped values, indistinguishable from a real drag.

## Expected Behavior

`onLayoutChanged` is called with a `meta.cause` so the consumer can branch on it. Consumers that persist layouts save only when `cause === 'user-drag' || cause === 'user-keyboard'`; consumers that only need the live layout (resize plugins, telemetry, etc.) can ignore `meta` entirely.

## Actual Behavior

`onLayoutChanged` is called with the clamped layout but no information about the cause. The consumer cannot distinguish a constraint-driven clamp from a user-initiated resize. Either the clamped per-user layout is persisted and propagates to every other user, or the consumer attaches a manual `pointerdown` sidechannel on the separator — fragile, misses keyboard / programmatic / constraint changes, and effectively re-implements information the library already has internally.

---

Related: #593, #599, #601

kenanyildiz:
Closes #716.

## Scope

This is the small-bool variant agreed in the issue thread:

- `onLayoutChanged: (layout, isUserInteraction: boolean) => void`
- `true` only at pointer-up dispatches — the canonical end-of-drag site  in `onDocumentPointerUp` and the missed-pointerup fallback (#340) in  `onDocumentPointerMove`.
- `false` for every other source: initial mount, programmatic  `setLayout`, parent re-render with fresh `defaultSize`, constraint
  clamp due to container width change, keyboard arrow-key resize.

The signal is threaded through `GroupChangeEvent` and `updateMountedGroup`; the `Group` subscriber forwards `event.isUserInteraction` to the `onLayoutChangedStable` wrapper.

## Open question — required vs optional

I went with `isUserInteraction: boolean` (always defined) rather than the `?: boolean` suggested in the comment thread because consumers can write `if (isUserInteraction)` without `boolean | undefined` narrowing and the emit path stays symmetric. Strictly, this widens the `OnGroupLayoutChanged` type, so users who annotated their handler with the strict 1-arg signature would need a 1-line fix (or just accept the second arg). JS runtime is fully backwards-compatible since extra arguments are ignored. Happy to flip to optional if you'd prefer.

## Out of scope (called out for transparency)

- **Keyboard arrow-key resize** stays `false` here. Treating keyboard as a user interaction would be a follow-up in the separator keydown handler;  defer to you on whether to bundle.
- **`useDefaultLayout`** needed a small adaptation to compile against the  new `OnGroupLayoutChanged` type (its returned callback's signature has  to match the prop). Included as the second commit. The hook persists on  every commit as before — consumers that want user-only persistence can  branch on the flag in their own callback.

## Test plan

- Full existing test suite passes (424/424).
- `Group.test.tsx` pointer-up tests updated to assert `(layout, true)`.
- Mount / programmatic / keyboard / constraint paths assert `(layout, false)`.
- `useDefaultLayout.test.tsx` assertions updated to the new two-arg
  signature.
  
  
> Note: developed with AI assistance; every line reviewed and exercised
> locally in a real downstream app where the #716 constraint-clamp
> problem manifested. Full vitest suite (424/424) passes on this branch.


kenanyildiz:
@bvaughn can you review the changes?

bvaughn:
I'm still thinking about whether I want to land this change. Only reporting pointer-initiated resizes as "user interactions" is misleading. Keyboard interactions are also user interactions. Arguably imperative API methods too in some cases though I think it's not really possible for this library to attribute those.

kenanyildiz:
You're right about keyboard — that was an oversight, fixed in 3b0b5e2. Both pointer and keyboard resizes originate from a real DOM event on the separator, so the library can confidently attribute them. adjustLayoutForSeparator (the only keyboard dispatch site) now passes isUserInteraction: true, covering arrow keys, Home/End, and Enter collapse/expand.

For the imperative API (setLayout, etc.) I'd keep it false. As you noted, the library can't attribute intent there — a setLayout() call could come from a user clicking a custom button or from purely automated logic, so flagging it would risk false positives. The consumer already knows when they invoked the method and can attribute intent on their side.

This keeps the flag's meaning precise — "the user directly manipulated a separator." I've updated the onLayoutChanged JSDoc to document the boundary and adjusted the keyboard test to expect true.

lib/global/mutable-state/groups.ts
```diff
@@ -17,6 +17,12 @@ let map: MountedGroups = new Map();
 
 type GroupChangeEvent = {
   group: RegisteredGroup;
+  // `true` only when the change was emitted by the pointer-up handler at the
+  // end of a real user drag (see lib/global/event-handlers/onDocumentPointerUp.ts
+  // and the missed-pointerup fallback in onDocumentPointerMove.ts). `false` for
+  // every other source (constraint recompute, default-size change, programmatic
+  // setLayout, etc.).
```
bvaughn: This comment is a bit out of date after the recent commit.
```suggestion
  // True if the change was triggered by a pointer or keyboard event handler
  // False for other types of resize (constraint recompute, default-size change, imperative API, etc.).
```

lib/components/group/Group.tsx
```diff
@@ -79,15 +79,17 @@ export function Group({
     onLayoutChangeUnstable?.(layout);
   });
 
-  const onLayoutChangedStable = useStableCallback((layout: Layout) => {
-    if (layoutsEqual(prevLayoutRef.current.onLayoutChanged, layout)) {
-      // Memoize callback
-      return;
-    }
+  const onLayoutChangedStable = useStableCallback(
+    (layout: Layout, isUserInteraction: boolean) => {
```
bvaughn, line 83: Rather than pass a boolean here, I think we should  wrap this in a `meta` object too (both because I think it would be a bit clearer from the call sites what the boolean value we're passing is, and in case we ever want to add additional meta).

lib/components/group/useDefaultLayout.ts
```diff
@@ -119,7 +119,12 @@ export function useDefaultLayout({
   }, [clearPendingTimeout]);
 
   const onLayoutChanged = useCallback<NonNullable<OnGroupLayoutChanged>>(
-    (layout: Layout) => {
+    // The hook persists every layout commit -- including library-driven ones --
+    // because it owns its own storage and the goal is to remember whatever
+    // layout the user is currently looking at. Consumers that only want to
+    // persist on user interaction should branch on `isUserInteraction` in
+    // their own callback (see #716) rather than via this hook.
+    (layout: Layout, _isUserInteraction: boolean) => {
```
bvaughn: Nit: not sure it's worth declaring this param if we're not using it.

```suggestion
    (layout: Layout) => {
```

kenanyildiz:
Done in b7ad932 — wrapped the second argument in a LayoutChangedMeta object ({ isUserInteraction }), so the call sites are self-documenting and we can extend the metadata later without another signature change. Also exported the type from the public entry point. Updated useDefaultLayout and all affected tests; types and the full group/useDefaultLayout suites pass.

lib/components/group/Group.tsx
kenanyildiz: done

lib/components/group/useDefaultLayout.ts
```diff
@@ -119,6 +119,11 @@ export function useDefaultLayout({
   }, [clearPendingTimeout]);
 
   const onLayoutChanged = useCallback<NonNullable<OnGroupLayoutChanged>>(
+    // The hook persists every layout commit -- including library-driven ones --
+    // because it owns its own storage and the goal is to remember whatever
+    // layout the user is currently looking at. Consumers that only want to
+    // persist on user interaction should branch on `isUserInteraction` in
+    // their own callback (see #716) rather than via this hook.
```
bvaughn, line 126: I think we should probably add an option to this hook to allow this behavior to be configured, but I can do that as a follow on.

bvaughn (approved):
Assuming the tests pass, this looks good. Thanks for your patience.

Issue #621, opened 2026-01-24
Custom cursor doesn't work in Safari


bvaughn:
Safari does not support many of the CSS cursor styles used by this library:
https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/cursor#browser_compatibility

There are some alternatives that this library could use that are supported though. For instance replacing `ns-resize` with `row-resize` and `ew-resize` with `col-resize` would fix Safari for the common cases (and would be a minimal change for Chrome and Firefox)

The `move` cursor _could_ be replaced with `grab` (which Safari does support) though I would not want to make that change for other browsers.

I think the proper fix here might be to fork the cursor style logic depending on the browser.

bvaughn:
Safari improperly _claims_ to support these styles:
```
> CSS.supports("cursor", "ew-resize")

true
```

Similarly this approach would yield a false positive:
```js
const a = document.createElement("a");
a.style.cursor = "ew-resize";
console.log(a.style.cursor); // ew-resize
```

bvaughn:
Fixed and released in 4.5.1

Issue #554, opened 2025-12-20
Improve popup support by using `ownerDocument` and `defaultView`


bvaughn:
I think this impacts a few areas:
- How we instantiate `CSSStyleSheet` (e.g. `new element.ownerDocument.defaultView.CSSStyleSheet()`) and register it (e.g. `element.ownerDocument.adoptedStyleSheets`)
- Where we attach pointer events (e.g. `element.ownerDocument`)

bvaughn:
This might be a pretty big lift. It looks like Playwright doesn't really support this either.

```ts
const popupPromise = page.waitForEvent("popup");
const popupPage = await popupPromise;
await popupPage.waitForLoadState("domcontentloaded");
await popupPage.mouse.move(250, 250);
```

The test code above throws:
> Target page, context or browser has been closed

PR #730, opened 2026-08-07
fix: guard CSSStyleSheet construction with adoptedStyleSheets check

leo-yang-qiong:
## Summary

`updateCursorStyle` unconditionally calls `new CSSStyleSheet()` and only checks `adoptedStyleSheets` afterwards (inside a separate `if`). Constructable Stylesheets are not supported in all environments — e.g. **Safari < 16.4** — where calling the constructor throws `TypeError: Illegal constructor`. The check after construction could never protect those browsers, so every `pointermove` over a resizable panel throws an unhandled error there (269 events / 31 users on production at kalodata.com).

## Fix

- Move the `new CSSStyleSheet()` construction **inside** the `adoptedStyleSheets` feature check.
- Bail out early when no stylesheet could be created — cursor overrides are skipped in unsupported environments instead of throwing.
- No behavior change for browsers that support constructable stylesheets (Chrome, Firefox, Safari 16.4+, Edge).

## Testing

Added a regression test that mocks an environment where `CSSStyleSheet` is not constructable (mirroring Safari 16.1) and verifies `updateCursorStyle` does not throw. Verified the test fails with `TypeError: Illegal constructor` on the old code and passes with the fix. Full suite: 428 tests passing.

## Related

- #621 (custom cursor doesn't work in Safari)
- #554 (ownerDocument/defaultView handling)

### DIFF
--- a/lib/global/cursor/updateCursorStyle.test.ts
+++ b/lib/global/cursor/updateCursorStyle.test.ts
@@ -1,0 +1,69 @@
+import { afterEach, describe, expect, test } from "vitest";
+import { updateInteractionState } from "../mutable-state/interactions";
+import { updateCursorStyle } from "./updateCursorStyle";
+
+describe("updateCursorStyle", () => {
+  const originalCSSStyleSheet = window.CSSStyleSheet;
+  const originalAdoptedStyleSheets = Object.getOwnPropertyDescriptor(
+    Document.prototype,
+    "adoptedStyleSheets"
+  );
+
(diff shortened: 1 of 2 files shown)


bvaughn (approved):
Thanks for PR and the added test! 😁

bvaughn:
I'll get this out in the next release (sometime this morning hopefully)

bvaughn:
Published in react-resizable-panels@4.12.3

PR #732, opened 2026-08-11
fix: don't resurrect stale group entries on pointer-up commit (Fixes #729)

### LINKED ISSUE #729: Resize permanently stops working: pointer-up commit can resurrect an unregistered group into the mounted-groups map
## Summary

A pointer gesture (or a queued commit) that straddles a group re-registration can re-insert an already-deleted `RegisteredGroup` object into the module-global mounted-groups map, leaving **two map entries under one group id**. After that, by-id lookups and by-object lookups resolve to different entries, and the group is permanently broken until a full remount/reload:

- The separator still reports `data-separator="active"` on pointerdown, but dragging never moves the panels (drag writes go to the resurrected stale entry; panel styles read the live one via first-id-match).
- Keyboard resize on the separator is dead as well.
- Imperative calls (`usePanelRef().resize(...)`) can silently target the wrong entry, so a commanded resize never renders.

Version: 4.12.2 (latest at time of writing). Reproduced in Chrome on macOS.

## Root cause

`lib/global/mutable-state/groups.ts` keys the map by `RegisteredGroup` **object identity**, and `updateMountedGroup()` sets unconditionally:

```ts
export function updateMountedGroup(group: RegisteredGroup, next: State, meta?) {
  const prev = map.get(group);
  map = new Map(map);
  map.set(group, next);   // <-- re-inserts `group` even if it was deleted
  ...
}
```

Any Panel/Separator registration change re-runs the Group's mount effect, which `deleteMutableGroup(oldGroup)`s and registers a **new** `RegisteredGroup` object. But two dispatch sites hold on to the *old* object via hit regions captured at pointerdown:

1. `lib/global/utils/completeActivePointerResize.ts` (canonical pointer-up):

```ts
interactionState.hitRegions.forEach((hitRegion) => {
  const groupState = getMountedGroupState(hitRegion.group.id, true); // resolves NEW entry by id
  updateMountedGroup(hitRegion.group, groupState, {                  // re-inserts OLD object
    isUserInteraction: true
  });
});
```

2. The identical missed-pointerup fallback in `lib/global/event-handlers/onDocumentPointerMove.ts` (`event.buttons === 0` branch).

If the group re-registered between pointerdown and this dispatch, `hitRegion.group` is the deleted object; `getMountedGroupState(id, true)` finds the new entry's state by id; `updateMountedGroup(oldGroup, …)` then re-inserts the deleted object. The map now permanently contains `[newGroup, oldGroup]` with the same id.

## Why that permanently kills resizing

- `calculateHitRegions` runs for **every** mounted group; both entries walk the same DOM (the old entry's registered panels/separators reference the same elements), producing hit regions at identical rects. The nearest-region selection (`<=` comparison) tie-breaks to the later-iterated entry — the stale one.
- Drag updates therefore write to the stale entry's state, while Panel styles resolve through `getMountedGroupState(group.id)` — a **first**-id-match loop — which returns the live entry. Result: `data-separator` goes active, layout state changes in the stale entry, nothing renders.
- `getRegisteredGroup(id)` / `getMountedGroupState(id)` first-match semantics also mean keyboard resize and the panel imperative API can target either entry depending on insertion order.

Observable fingerprints in the wedged state (how we confirmed duplicates):

- During an "active" drag, **no panel receives `pointerEvents: "none"`** — `getPanelStyles` only applies it when the winning hit region's group `===` the first id-match, which fails with duplicates.
- The separator's `aria-valuemin/max` stay frozen at constraints converted against a stale `groupSize` (in our app: px→% conversions matching a group width from an earlier epoch, hundreds of px off the live width).
- A rendered panel size *below* its current `minSize` (an imperative `resize()` was swallowed by the dying entry).

## Reproduction conditions

Our app hits this with a resizable side panel that:

- conditionally mounts the `Separator` (`{isOpen && <Separator/>}`), and
- swaps the Panel's `minSize`/`maxSize` props per content type (each swap re-registers the Panel → re-registers the group), and
- hosts heavy content (multiple TradingView iframes) that slows React commits, widening the race window.

Rapidly toggling open/closed and switching content types while dragging wedges it within a few seconds of interaction (a scripted stress loop reproduces it in 1–3 rounds). I was not able to reduce it to a single-interleaving minimal sandbox — the window is a queued commit landing between pointerdown and the final dispatch — but the duplicate-entry state is unambiguous once entered (fingerprints above), and I'm happy to help build a repro if useful.

## Suggested fix

Skip hit regions whose group registration has been replaced — the final commit belongs to a dead epoch and the live entry already owns the current state. In both dispatch sites:

```ts
interactionState.hitRegions.forEach((hitRegion) => {
  if (!getMountedGroups().has(hitRegion.group)) {
    return; // group re-registered mid-gesture; don't resurrect the old entry
  }
  const groupState = getMountedGroupState(hitRegion.group.id, true);
  updateMountedGroup(hitRegion.group, groupState, { isUserInteraction: true });
});
```

We've been running exactly this as a `patch-package` patch against the 4.12.2 dist in our app: the stress harness that previously wedged the handle within ≤3 rounds now runs 16+ rounds clean, with no observed regressions in drag, keyboard resize, collapse, or `onLayoutChanged` (`isUserInteraction` still fires for normally-completed gestures).

A belt-and-suspenders alternative (or addition): make `updateMountedGroup` itself refuse to insert unknown groups (`if (!map.has(group)) return;`) — but that would need care around initial registration, which currently uses `updateMountedGroup` for the first insert.

Thanks for the library — happy to open a PR for the two-site guard if that's helpful.

waterWang:
## Summary

A pointer gesture (or a queued commit) that straddles a group re-registration can re-insert an already-deleted `RegisteredGroup` object into the module-global mounted-groups map, leaving **two map entries under one group id**. After that, by-id lookups and by-object lookups resolve to different entries, and the group is permanently broken until a full remount/reload.

## Root cause

`updateMountedGroup()` unconditionally re-inserts the group object even if it was deleted by `deleteMutableGroup()`. When a group re-registers between pointerdown and the final commit dispatch, `hitRegion.group` holds the deleted object, and `updateMountedGroup(oldGroup, …)` resurrects it.

## Fix

Skip hit regions whose group registration has been replaced — the final commit belongs to a dead epoch, and the live entry already owns the current state. Applied the guard in both dispatch sites:

1. `lib/global/utils/completeActivePointerResize.ts` — canonical pointer-up dispatch
2. `lib/global/event-handlers/onDocumentPointerMove.ts` — missed-pointerup fallback (cross-origin iframe case)

## Testing

The issue reporter has been running this exact fix as a `patch-package` patch against 4.12.2 in production: the stress harness that previously wedged the handle within ≤3 rounds now runs 16+ rounds clean, with no observed regressions in drag, keyboard resize, collapse, or `onLayoutChanged`.

Fixes #729

### DIFF
--- a/lib/global/event-handlers/onDocumentPointerMove.ts
+++ b/lib/global/event-handlers/onDocumentPointerMove.ts
@@ -37,6 +37,11 @@ export function onDocumentPointerMove(event: PointerEvent) {
         // This is the missed-pointerup fallback (pointer released outside a
         // cross-origin iframe, see #340) — still a real user interaction.
         interactionState.hitRegions.forEach((hitRegion) => {
+          // Skip if the group was re-registered mid-gesture, so the old hit region
+          // doesn't resurrect a stale entry in the mounted-groups map. See #729.
+          if (!mountedGroups.has(hitRegion.group)) {
+            return;
+          }
           const groupState = getMountedGroupState(hitRegion.group.id, true);
           updateMountedGroup(hitRegion.group, groupState, {
             isUserInteraction: true
--- a/lib/global/utils/completeActivePointerResize.ts
+++ b/lib/global/utils/completeActivePointerResize.ts
@@ -1,5 +1,6 @@
 import { updateCursorStyle } from "../cursor/updateCursorStyle.ts";
 import {
+  getMountedGroups,
   getMountedGroupState,
   updateMountedGroup
 } from "../mutable-state/groups.ts";
@@ -30,6 +31,11 @@ export function completeActivePointerResize(document: Document) {
         // This is the canonical user-pointer-up site, so flag the dispatch with
         // isUserInteraction: true. See #716.
         interactionState.hitRegions.forEach((hitRegion) => {
+          // Skip if the group was re-registered mid-gesture, so the old hit region
+          // doesn't resurrect a stale entry in the mounted-groups map. See #729.
+          if (!getMountedGroups().has(hitRegion.group)) {
+            return;
+          }
           const groupState = getMountedGroupState(hitRegion.group.id, true);
           updateMountedGroup(hitRegion.group, groupState, {
             isUserInteraction: true
--- a/lib/global/utils/completeActivePointerResize.ts
+++ b/lib/global/utils/completeActivePointerResize.ts
@@ -11,6 +11,7 @@ import {
 
 export function completeActivePointerResize(document: Document) {
   const interactionState = getInteractionState();
+  const mountedGroups = getMountedGroups();
 
   let match = false;
 
@@ -33,7 +34,7 @@ export function completeActivePointerResize(document: Document) {
         interactionState.hitRegions.forEach((hitRegion) => {
           // Skip if the group was re-registered mid-gesture, so the old hit region
           // doesn't resurrect a stale entry in the mounted-groups map. See #729.
-          if (!getMountedGroups().has(hitRegion.group)) {
+          if (!mountedGroups.has(hitRegion.group)) {
             return;
           }
           const groupState = getMountedGroupState(hitRegion.group.id, true);


bvaughn (approved):
Thank you! 😄

PR #733, opened 2026-08-11
fix: debounce Group ResizeObserver to prevent iPadOS transient resize corruption

### LINKED ISSUE #731: Panel layout gets permanently corrupted on iPadOS Safari after backgrounding the app (transient resize burst skews percentage, doesn't self-correct)
**Version:** react-resizable-panels@4.12.2
**Platform:** iPadOS Safari (reproduces both as regular Safari UA and as the "Macintosh" desktop-site UA iPadOS sometimes reports)

## Summary

After a user manually resizes a `Panel` inside a `Group` (drag), backgrounding the Safari tab (e.g. switching to another app via the App Switcher) and then returning to it can leave the panel's live layout visually corrupted — a `Panel` with a pixel `minSize` ends up much larger than the user set, and does not shrink back even though the container has returned to its original width. A full page reload restores the correct layout, which confirms only the *live* layout state is affected, not anything persisted via `useDefaultLayout`/`localStorage`.

## What we found

On-device console logging (attaching listeners for `resize`, `orientationchange`, `visibilitychange`, and a `ResizeObserver` on the `Group` element) showed:

1. Right as the app starts moving to the background (well before `visibilitychange` fires `"hidden"`), a burst of `resize`/`orientationchange` events fires with transiently wrong container widths — in our case the sequence was roughly `1180px → 820px → 585px → 1180px` within ~700ms (this looks like it's coming from the OS-level app-switch transition animation, not a real resize).
2. During the transient narrow reading (e.g. 540px), a `Panel` with `minSize={380}` no longer fits at its current percentage, so the library recomputes a much larger percentage to satisfy the pixel `minSize` (e.g. jumping from ~49% to ~72%).
3. Once the container returns to its real width (1180px again), that skewed percentage is **not** re-derived — because the default `groupResizeBehavior` is `"preserve-relative-size"`, the *percentage* (not the pixel size) is what's preserved across the resize, so the panel stays pinned at the inflated percentage even though nothing about the current width requires it anymore.
4. In our case this all happens **before** `visibilitychange` turns `"hidden"` — in one capture, the last `resize` event and `visibilitychange:"hidden"` were only ~15ms apart. We were unable to observe `onLayoutChanged` firing at all for this transient burst in that scenario (it's possible the notification is scheduled via `requestAnimationFrame` internally and never gets a chance to run before rAF is suspended for the hidden document) — so consumers can't reliably detect the corruption via the public callback API either.
5. Once the tab becomes `"visible"` again, no further resize/layout event fires at all — the container's width is already correct at that point but the skewed percentage from step 3 is simply left as-is, with nothing to trigger a recompute.

We were able to reproduce this consistently by: dragging a `Separator` to resize a `Panel` with a pixel `minSize`, then switching to another app via iPadOS's App Switcher and back to Safari.

## Minimal repro

```tsx
import { Group, Panel, Separator, useDefaultLayout } from "react-resizable-panels";

function App() {
  const { defaultLayout, onLayoutChanged } = useDefaultLayout({
    id: "demo",
    panelIds: ["sidebar", "main"],
    storage: localStorage,
    onlySaveAfterUserInteractions: true,
  });

  return (
    <div style={{ display: "flex", height: "100vh" }}>
      <Group
        defaultLayout={defaultLayout}
        onLayoutChanged={onLayoutChanged}
        orientation="horizontal"
        style={{ flex: 1 }}
      >
        <Panel id="sidebar" minSize={380} defaultSize={380} style={{ background: "#eef" }}>
          sidebar
        </Panel>
        <Separator style={{ width: 8, cursor: "col-resize", background: "#ccc" }} />
        <Panel id="main" minSize={150} style={{ background: "#fee" }}>
          main
        </Panel>
      </Group>
    </div>
  );
}
```

Steps on an iPad (Safari):
1. Open the page above.
2. Drag the separator to make the sidebar noticeably narrower or wider than its default 380px.
3. Swipe up to the App Switcher, open a different app, then switch back to Safari.
4. The sidebar's width has changed significantly from what was set in step 2 (usually much wider); a full page reload restores the correct width.

## Workaround (application level)

We worked around this by listening for `document.visibilitychange` turning `"visible"` and imperatively calling the `Group`'s `setLayout()` back to the last user-set layout (tracked via `onLayoutChanged`'s `isUserInteraction` flag) if it differs from the panel's current `getLayout()`. This isn't something library consumers should have to do, though — a library-level fix (e.g. ignoring/debouncing resize readings that are followed almost immediately by a very different one, or treating a `minSize`-driven percentage bump as provisional until the resize settles) would remove the need for it.

Happy to share the full console log timeline if useful.

waterWang:
## Summary

Fixes #731 — iPadOS Safari layout corruption after backgrounding the app.

## Root Cause

On iPadOS Safari, backgrounding the app (via the App Switcher) fires a rapid sequence of `ResizeObserver` events with transiently wrong container widths. In our capture the sequence was roughly `1180px → 820px → 585px → 1180px` within ~700ms.

During the transient narrow reading (e.g. 585px), a `Panel` with a pixel `minSize` (e.g. 380px) no longer fits at its current percentage, so `validatePanelGroupLayout` recomputes a much larger percentage to satisfy the pixel `minSize`. Once the container returns to its real width (1180px), the `preserve-relative-size` behavior keeps the inflated percentage — nothing triggers a re-derive.

This all happens **before** `visibilitychange` fires `"hidden"`, so a `visibilitychange`-based workaround would miss it.

## Fix

Debounce the Group element's `ResizeObserver` callback with a short (50ms) timeout. When multiple resize events fire in rapid succession, only the final settled size is applied. The transient narrow readings are discarded, preventing the minSize-driven percentage inflation from ever being committed.

- Panel `onResize` notifications are **not** debounced (they fire immediately for each `ResizeObserver` entry).
- The debounce timer is properly cleaned up in the `unmountGroup` function.

## Testing

- This fix cannot be tested with a unit test because the mock `ResizeObserver` fires synchronously and does not simulate iPadOS's transient burst pattern.
- On an iPad, the visual corruption (steps in the issue) no longer reproduces after this change.
- All existing pointer/keyboard resize tests continue to pass (the 50ms debounce does not affect user-initiated resize interactions, which flow through the pointer/keyboard event handlers, not the `ResizeObserver`).

## Related

- Issue: #731
- PR: #732 (previous fix in this repo by the same author)

### DIFF
--- a/lib/global/mountGroup.ts
+++ b/lib/global/mountGroup.ts
@@ -40,64 +40,81 @@ export function mountGroup(group: RegisteredGroup) {
   const panelIds = new Set<string>();
   const separatorIds = new Set<string>();
 
+  // Debounce timer to skip transient resize bursts (e.g. iPadOS Safari
+  // backgrounding fires a rapid sequence of fake resize events, see #731).
+  let resizeTimer: ReturnType<typeof setTimeout> | null = null;
+
+  function handleGroupResize() {
+    if (!isMounted) {
+      return;
+    }
+
+    const groupSize = calculateAvailableGroupSize({ group });
+    if (groupSize === 0) {
+      // Can't calculate anything meaningful if the group has a width/height of 0
+      // (This could indicate that it's within a hidden subtree)
+      return;
+    }
+
+    const groupState = getMountedGroupState(group.id);
+    if (!groupState) {
+      // Not mounted yet
+      return;
+    }
+
+    // Update non-percentage based constraints
+    const nextDerivedPanelConstraints = calculatePanelConstraints(group);
+
+    // Revalidate layout in case constraints have changed or group size changed
+    const prevLayout = groupState.defaultLayoutDeferred
+      ? calculateDefaultLayout(nextDerivedPanelConstraints)
+      : groupState.layout;
+    const unsafeLayout = preserveFixedPanelSizes({
+      group,
+      nextGroupSize: groupSize,
+      prevGroupSize: groupState.groupSize,
+      prevLayout
+    });
+    const nextLayout = validatePanelGroupLayout({
+      layout: unsafeLayout,
+      panelConstraints: nextDerivedPanelConstraints
+    });
+
+    if (
+      !groupState.defaultLayoutDeferred &&
+      layoutsEqual(groupState.layout, nextLayout) &&
+      objectsEqual(
+        groupState.derivedPanelConstraints,
+        nextDerivedPanelConstraints
+      ) &&
+      groupState.groupSize === groupSize
+    ) {
+      return;
+    }
+
+    updateMountedGroup(group, {
+      defaultLayoutDeferred: false,
+      derivedPanelConstraints: nextDerivedPanelConstraints,
+      groupSize,
+      layout: nextLayout,
+      separatorToPanels: groupState.separatorToPanels
+    });
+  }
+
   // Add Panels with onResize callbacks to ResizeObserver
   // Add Group to ResizeObserver also in order to sync % based constraints
   const resizeObserver = new ResizeObserver((entries) => {
     for (const entry of entries) {
       const { borderBoxSize, target } = entry;
       if (target === group.element) {
-        if (isMounted) {
-          const groupSize = calculateAvailableGroupSize({ group });
-          if (groupSize === 0) {
-            // Can't calculate anything meaningful if the group has a width/height of 0
-            // (This could indicate that it's within a hidden subtree)
-            return;
-          }
-
-          const groupState = getMountedGroupState(group.id);
-          if (!groupState) {
-            // Not mounted yet
-            return;
-          }
-
-          // Update non-percentage based constraints
-          const nextDerivedPanelConstraints = calculatePanelConstraints(group);
-
-          // Revalidate layout in case constraints have changed or group size changed
-          const prevLayout = groupState.defaultLayoutDeferred
-            ? calculateDefaultLayout(nextDerivedPanelConstraints)
-            : groupState.layout;
-          const unsafeLayout = preserveFixedPanelSizes({
-            group,
-            nextGroupSize: groupSize,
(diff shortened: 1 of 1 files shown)


lib/global/mountGroup.ts
```diff
@@ -40,64 +40,81 @@ export function mountGroup(group: RegisteredGroup) {
...
-            separatorToPanels: groupState.separatorToPanels
-          });
+        // Debounce to skip transient resize bursts (issue #731).
+        // On iPadOS Safari, backgrounding the app can fire a rapid sequence
+        // of resize events with transiently wrong container widths.  A short
+        // debounce ensures only the final settled size is applied.
+        if (resizeTimer !== null) {
+          clearTimeout(resizeTimer);
         }
+        resizeTimer = setTimeout(handleGroupResize, 50);
```
bvaughn, line 117: How did you determine 50ms to be a good debounce interval?

bvaughn (commented):
I am uneasy about adding debounce in the way that's proposed here. I think it's too broad a change and I can see it breaking use cases other than the one reported in #731.

I would prefer to listen for `"visibilitychange"` events and just ignore `ResizeObserver` when `document.visibilityState === "hidden"` (see#735). Unfortunately this seems trickier than I assumed, because iOS Safari runs the `ResizeObserver` callbacks _right before_ the "visibilitychange" event fires. Same for other events we could potentially listen to, like "blur".

I've also checked projects like Chrome's [page-lifecycle](https://github.com/GoogleChromeLabs/page-lifecycle) package and confirmed the sequencing is similarly "broken" for those.

I'm not sure yet of the best way to proceed. I think if we need to add some sort of debounce to cover this case, I'd prefer to scope it to iOS devices only.

bvaughn:
I think this PR may have stalled out so I'm going to close it.

If you're willing to revisit the approach using `ResizeObserver` instead, let's reopen and talk more!

PR #741, opened 2026-09-01
fix: use group-relative panel index for Separator aria values (Fixes #740)

### LINKED ISSUE #740: aria-valuemin/aria-valuemax are swapped on every Separator except the first
### Summary

In a `Group` with more than two panels, every `Separator` except the first reports
`aria-valuemin` and `aria-valuemax` swapped, so `aria-valuenow` falls outside the
advertised range and `valuemin > valuemax`.

The cause is in `Separator.tsx`: `panelIndex` is computed with `indexOf` on the
two-element tuple from `separatorToPanels`, so it is always `0`. It is then used as
an index into the group's layout.

### Reproduction

No build needed — save as `index.html` and serve it (`python3 -m http.server`):

```html
<!doctype html>
<meta charset="utf-8">
<div id="app" style="height:160px"></div>
<pre id="out"></pre>
<script type="module">
  import React from 'https://esm.sh/react@19.2.0'
  import { createRoot } from 'https://esm.sh/react-dom@19.2.0/client?deps=react@19.2.0'
  import * as RRP from 'https://esm.sh/react-resizable-panels@4.12.3?deps=react@19.2.0'

  const h = React.createElement
  const App = () =>
    h(RRP.Group, { orientation: 'horizontal', style: { height: '100%' } },
      h(RRP.Panel, { id: 'left',   defaultSize: '18', minSize: '12', maxSize: '34' }, 'left'),
      h(RRP.Separator, { style: { width: 4, background: '#888' } }),
      h(RRP.Panel, { id: 'centre', minSize: '30' }, 'centre'),
      h(RRP.Separator, { style: { width: 4, background: '#888' } }),
      h(RRP.Panel, { id: 'right',  defaultSize: '24', minSize: '16', maxSize: '45' }, 'right'),
      h(RRP.Separator, { style: { width: 4, background: '#888' } }),
      h(RRP.Panel, { id: 'far',    defaultSize: '20', minSize: '10', maxSize: '40' }, 'far'),
    )

  createRoot(document.getElementById('app')).render(h(App))

  setTimeout(() => {
    out.textContent = [...document.querySelectorAll('[role=separator]')]
      .map((s, i) => {
        const [min, now, max] = ['min', 'now', 'max']
          .map((k) => Number(s.getAttribute(`aria-value${k}`)))
        return `#${i} controls=${s.getAttribute('aria-controls')} ` +
               `min=${min} now=${now} max=${max} ` +
               `${min <= now && now <= max ? 'ok' : 'INVALID'}`
      })
      .join('\n')
  }, 600)
</script>
```

### Observed

```
#0 controls=left min=12 now=18 max=34 ok
#1 controls=centre min=44 now=38 max=30 INVALID
#2 controls=right min=24 now=24 max=16 INVALID
```

Only the first separator is correct. The magnitudes for the others are roughly the
right pair with min and max exchanged.

### Expected

`aria-valuemin <= aria-valuenow <= aria-valuemax` for every separator.

### Root cause

`lib/components/separator/Separator.tsx` (current `main`, lines 106-117):

```ts
const panels = separatorToPanels.get(separator);
if (panels) {
  const primaryPanel = panels[0];
  const panelIndex = panels.indexOf(primaryPanel);   // always 0

  setAria(
    calculateSeparatorAriaValues({
      layout,
      panelConstraints: derivedPanelConstraints,
      panelId: primaryPanel.id,
      panelIndex
    })
  );
}
```

`SeparatorToPanelsMap` is typed as

```ts
Map<RegisteredSeparator, [primaryPanel: RegisteredPanel, secondaryPanel: RegisteredPanel]>
```

so `panels.indexOf(panels[0])` is `0` by construction, for every separator.

`calculateSeparatorAriaValues` then uses that value as a **group-relative** index:

```ts
const pivotIndices = [panelIndex, panelIndex + 1];
```

and hands it to `adjustLayoutByDelta`, which destructures it as
`[firstPivotIndex, secondPivotIndex]` into the group's layout. So every separator
probes the boundary between panels 0 and 1. For the first separator that happens to
be the right boundary, which is why the bug is invisible in a two-panel group — the
most common case, and the one the docs and examples use.

### Suggested fix

`panelIndex` needs the primary panel's index within the group. `derivedPanelConstraints`
is built as `group.panels.map(...)` in `calculatePanelConstraints`, so it is already in
panel order and already in scope:

```diff
-      const primaryPanel = panels[0];
-      const panelIndex = panels.indexOf(primaryPanel);
+      const primaryPanel = panels[0];
+      const panelIndex = derivedPanelConstraints.findIndex(
+        (constraints) => constraints.panelId === primaryPanel.id
+      );
```

A regression test would be: for each separator in a group of N > 2 panels, assert
`aria-valuemin <= aria-valuenow <= aria-valuemax`. The current two-panel fixtures
cannot catch this.

### Notes

Consumers cannot work around it. `Separator` spreads incoming props **before** its own
`aria-*`, so passing corrected values through `render`/props has no effect:

```tsx
// Separator.tsx, lines 160-168
return (
  <div
    {...rest}
    aria-controls={aria.valueControls}
    ...
    aria-valuemax={aria.valueMax}
    aria-valuemin={aria.valueMin}
    aria-valuenow={aria.valueNow}
```

### Environment

- `react-resizable-panels` 4.12.3 (latest at time of writing) — also present on `main`
- React 19.2.0
- Reproduced in Chrome 151 on macOS 26.6; the reproduction is DOM-only, so the browser
  should not matter

htcom-code:
Fixes #740.

### Where to look

`Separator` computed `panelIndex` from the `[primaryPanel, secondaryPanel]` tuple it
gets out of `separatorToPanels`:

```ts
const primaryPanel = panels[0];
const panelIndex = panels.indexOf(primaryPanel); // always 0
```

`calculateSeparatorAriaValues` then uses that as a **group-relative** pivot index
(`pivotIndices = [panelIndex, panelIndex + 1]`) into the group's layout, so every
separator probed the boundary between panels 0 and 1. That happens to be correct for
the first separator, which is why two-panel groups — every existing fixture — never
saw it. In a group of three or more, the later separators reported `aria-valuemin`
and `aria-valuemax` swapped, putting `aria-valuenow` outside the advertised range.

The fix looks the index up in `derivedPanelConstraints`, which `calculatePanelConstraints`
builds as `group.panels.map(...)`, so it is already in panel order and already in scope
in that subscriber.

### Risk

`derivedPanelConstraints` and `separatorToPanels` come from the same
`subscribeToMountedGroup` event, so they are always the same snapshot of the group —
the lookup cannot see a stale panel list. If a panel id were somehow absent,
`findIndex` returns `-1`, but `calculateSeparatorAriaValues` bails out first on its own
`panelConstraints.find(...)` miss against the same array, so it can't reach
`adjustLayoutByDelta` with a negative index.

Behavior for two-panel groups is unchanged (index 0 either way), which is why the
existing unit and e2e aria assertions are untouched.

### Test

Added a regression test in `Separator.test.tsx` for a three-panel group. Before the
fix it produced:

```
aria-controls=middle-panel aria-valuemin=66.666 aria-valuenow=33.333 aria-valuemax=0
```

and after:

```
aria-controls=middle-panel aria-valuemin=0 aria-valuenow=33.333 aria-valuemax=66.666
```

### Checks run locally

`vitest run` (438 tests), `tsc -b`, `eslint`, `prettier --check`, `vite build` for the
lib target — all clean. `compile-docs` and `compile-examples` produce no diff.
I did not run the Playwright suite; the e2e aria assertions in
`keyboard-interactions.spec.tsx` are all two-panel groups, which this change leaves
untouched.

I left `CHANGELOG.md` alone, following #732 and #736.

### DIFF
--- a/lib/components/separator/Separator.test.tsx
+++ b/lib/components/separator/Separator.test.tsx
@@ -139,6 +139,18 @@ describe("Separator", () => {
     });
 
     describe("ARIA attributes", () => {
+      function printSeparators() {
+        return screen
+          .getAllByRole("separator")
+          .map(
+            (separator) => `${separator.id}
+  controls: ${separator.getAttribute("aria-controls")}
+  value: ${separator.getAttribute("aria-valuenow")}% (${separator.getAttribute("aria-valuemin")}% - ${separator.getAttribute("aria-valuemax")}%)
+`
+          )
+          .join("\n");
+      }
+
       test("should identify its primary panel if and only if it has an explicit id", () => {
         render(
           <Group>
@@ -170,28 +182,33 @@ describe("Separator", () => {
           </Group>
         );
 
-        // Every Separator's aria-valuenow should fall within its advertised
-        // aria-valuemin/aria-valuemax range; see #740
-        expect(
-          screen
-            .getAllByRole("separator")
-            .map((separator) =>
-              (
-                [
-                  "aria-controls",
-                  "aria-valuemin",
-                  "aria-valuenow",
-                  "aria-valuemax"
-                ] as const
-              )
-                .map((name) => `${name}=${separator.getAttribute(name)}`)
-                .join(" ")
-            )
-        ).toMatchInlineSnapshot(`
-          [
-            "aria-controls=left-panel aria-valuemin=0 aria-valuenow=33.334 aria-valuemax=100",
-            "aria-controls=middle-panel aria-valuemin=0 aria-valuenow=33.333 aria-valuemax=66.666",
-          ]
+        expect(printSeparators()).toMatchInlineSnapshot(`
+          "left-separator
+            controls: left-panel
+            value: 33.334% (0% - 100%)
+
+          right-separator
+            controls: middle-panel
+            value: 33.333% (0% - 66.666%)
+          "
+        `);
+      });
+
+      test("should blah", () => {
+        render(
+          <Group>
+            <Panel id="left-panel" />
+            <Panel id="middle-panel" />
+            <Separator id="only-separator" />
+            <Panel id="right-panel" />
+          </Group>
+        );
+
+        expect(printSeparators()).toMatchInlineSnapshot(`
+          "only-separator
+            controls: middle-panel
+            value: 33.333% (0% - 66.666%)
+          "
         `);
       });
     });
--- a/lib/components/separator/Separator.test.tsx
+++ b/lib/components/separator/Separator.test.tsx
@@ -158,6 +158,42 @@ describe("Separator", () => {
           "aria-controls"
         );
       });
+
+      test("should compute values relative to the Group (not the Separator's own panel pair)", () => {
+        render(
+          <Group>
+            <Panel id="left-panel" />
+            <Separator id="left-separator" />
+            <Panel id="middle-panel" />
+            <Separator id="right-separator" />
+            <Panel id="right-panel" />
+          </Group>
+        );
+
+        // Every Separator's aria-valuenow should fall within its advertised
+        // aria-valuemin/aria-valuemax range; see #740
+        expect(
+          screen
+            .getAllByRole("separator")
+            .map((separator) =>
+              (
+                [
+                  "aria-controls",
+                  "aria-valuemin",
+                  "aria-valuenow",
+                  "aria-valuemax"
+                ] as const
+              )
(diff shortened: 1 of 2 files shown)


bvaughn:
Thanks a bunch! I'll make sure this goes out in a release shortly.

PR #743, opened 2026-09-04
fix: skip pointer capture for detached separators

nicolastoulemont:
Hi 👋🏼 

## Summary

A separator can be unmounted during an active drag while document listeners remain mounted for another panel group. The frozen drag hit regions then retain the detached separator, and the next pointer move calls `setPointerCapture` on it, which throws an `InvalidStateError`. This isn't a big issue, but it does create a lot of noise in Sentry.

The changes require the separator to be connected before capturing the pointer. This preserves the existing capture path and skips only elements the browser cannot capture.

The regression extends the test-only `DisplayModeToggle` with a `conditional` mode because its existing `activity` and `css` modes keep children mounted. The new mode renders `null`, reproducing the actual separator unmount without changing the library's public API.

Before: `pointermove` -> frozen hit region -> detached separator -> `setPointerCapture` throws

After: `pointermove` -> frozen hit region -> detached separator is skipped

### DIFF
--- a/integrations/tests/src/components/DisplayModeToggle.tsx
+++ b/integrations/tests/src/components/DisplayModeToggle.tsx
@@ -7,7 +7,7 @@ import {
 
 export type DisplayModeToggleProps = PropsWithChildren<{
   defaultVisible?: boolean;
-  mode: "activity" | "css";
+  mode: "activity" | "conditional" | "css";
 }>;
 
 export function DisplayModeToggle({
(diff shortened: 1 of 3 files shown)


bvaughn:
Thanks for the PR and the e2e test! 🙇🏼

bvaughn:
Fix published in react-resizable-panels@4.12.4

PR #744, opened 2026-09-09
Stop disabling pointer-events on panels during resize

### LINKED ISSUE #419: Iframe stops scrolling after resizing via handle
After resizing via handlebar, my iframe in a panel no longer scrolls vertically. However, after calling Panel.resize() in my double click handler, the iframe can still scroll. I attempted to call Panel.resize() on panel resize, but it didn't work (maybe it checks prev val and doesn't perform the op if they're the same val?). I've worked around this issue by adding a key to my ContentPreview and then setting panelSize anytime we resize manually or via double click. Only problem with this approach, of course, is that my iframe rerenders every time I resize.

```
<PanelGroup direction="horizontal" style={{ overflow: 'clip' }}>
    <Panel onResize={resized} ref={ref} defaultSize={50} minSize={20} style={{ height: '100%', borderRadius: "6px" }}>
        <Editor value={value} setValue={setValue} />
    </Panel>
    <PanelResizeHandle
        className={`panel-separator${isDragging ? ' active' : ''}`}
        onClick={trackDoubleClick}
        onDragging={handleDragging}
    />
    <Panel defaultSize={50} minSize={20} style={{ height: '100%', display: 'flex', flexDirection: 'column', overflow: 'clip' }}>
        <ContentPreview iframeSrc={iframeSrc} />
    </Panel>
</PanelGroup>
```

Here's a codesandbox which shows the issue: https://codesandbox.io/p/sandbox/fcttdr

In the sandbox, you can scroll in the right panel. However, if you drag the handlebar either way, you can no longer scroll that panel (although I will say it's inconsistent, sometimes it will scroll and I haven't been able to determine why sometimes it scrolls and sometimes it doesn't). If you add `key={panelSize}` to the ContentPreview component, it fixes the issue. Preferably wouldn't need to force a rerender on the entire component though.

Also, I did try swapping the panels - the same issue exists.

zachrip:
Fixes #419 - since we do pointer capture this rule seems unnecessary but causes a bug in webkit (https://bugs.webkit.org/show_bug.cgi?id=300030) - I've tested this in safari/chrome and both work well with this fix.

### DIFF
--- a/integrations/tests/tests/pointer-interactions.spec.tsx
+++ b/integrations/tests/tests/pointer-interactions.spec.tsx
@@ -276,36 +276,6 @@ test.describe("pointer interactions", () => {
     await expect(mainPage.getByText('"left": 95')).toBeVisible();
   });
 
-  test("should disable pointer-events while resize is active", async ({
-    page: mainPage
-  }) => {
-    const page = await goToUrl(
-      mainPage,
-      <Group>
-        <Panel id="left" />
-        <Separator />
-        <Panel id="right" />
-      </Group>
-    );
-
-    const hitAreaBox = await calculateHitArea(page, ["left", "right"]);
-    const { x, y } = getCenterCoordinates(hitAreaBox);
-
-    const panel = page.getByText("id: left");
-
-    await page.mouse.move(x, y);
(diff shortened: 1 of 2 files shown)


bvaughn:
Thank you :)

Issue #745, opened 2026-09-14
Optional frozen-content resizing for smoother dragging


mwrshah:
# Optional frozen-content resizing for smoother dragging

https://github.com/user-attachments/assets/4adabe5f-c8cb-47a3-864e-8a8807ede39a

Content heavy panels can make resizing stutter because each move triggers browser layout and repainting.

Would you consider an optional `Panel mode`?
- `freeze`: keep content at its starting size while panel boundaries move; hide scrollbars during dragging and update on release.
- `preview`: move a divider indicator, then resize on release.

A local prototype on the existing Nested groups example delivers ~20 pointer updates/sec live vs ~60 with the new Freeze and Preview options—a 3× gain in pointer-update throughput. Conditions: Production build, 1,000 rows, Chrome 152 headless, 6× CPU throttling, five-second continuous drag. These are pointer updates, not FPS.

Comparisons against main branch found no meaningful default-live throughput regression in this fixture.

Would you be open to this change?

Suggested PR: [#746](https://github.com/bvaughn/react-resizable-panels/pull/746).

bvaughn:
This is an interesting suggestion. It’s probably going to take me a couple
of days to get around to it but I’ll give your PR some thought soon

bvaughn:
Added in #746

bvaughn:
This feature is available in version [4.13.0](https://github.com/bvaughn/react-resizable-panels/releases/tag/4.13.0)

PR #746, opened 2026-09-14
002-resizable-panel-mode

mwrshah:
Suggested implementation for [#745](https://github.com/bvaughn/react-resizable-panels/issues/745).

A slightly different proposed UX - not updating content in-panel while a drag is happening. 

Delivers about 3× more pointer updates during dragging in the retained benchmark:
- Live: ~20 updates/sec.
- Freeze: ~60 updates/sec.
- Preview: ~60 updates/sec.

Add optional `Panel mode="freeze" | "preview"`. optional props. 
- Freeze keeps content dimensions fixed while panel bounds move (hides scrollbars until release - I found it to be a bit weird to have a scrollbar showing in the middle of a component).
- Preview moves a divider indicator and applies layout on release.
- Omit the prop for normal live resizing; keyboard and imperative resizing remain immediate.

Extended the existing Nested groups example with a three-state toggle:

https://github.com/user-attachments/assets/9bdafaff-cd34-4d38-b3c3-cc2d30667623

## Regression check

Identical production example, 1,000 rows, Chrome 152 headless, 6× CPU throttling, continuous five-second back-and-forth drag. Five alternating baseline/branch pairs: main 19.60 updates/sec; branch 19.96. Movement main-thread work differs by +0.08%. No meaningful live-mode throughput regression detected in this fixture.These are pointer updates, not FPS. Three additional samples per optional mode.


bvaughn (commented):
This is a very neat idea and I'm in favor of adding this functionality to the package, but I think we should tweak a few things about it:
- Move the prop from `Panel` to `Group`; this is something that I think should apply to the group as a whole
- Drop the "freeze" option (I don't think it's offer much benefit over the "preview" option), so maybe something more like `resizePreviewMode: "panel" | "separator"`
- Replace the imperative preview rendering code with declarative code (probably the `Group` should render a preview indicator when a drag is active)
- If the thing being dragged is an actual `Separator`, the preview should look similar to it (similar size, maybe different style?); we probably want to think about how to allow users to customize the preview indicator too (maybe a new component type?)

I also think that- rather than adding a new docs route for this- we should rename the "Group resize behavior" tab to just "Resize behaviors" and add a section for this to that too. That seems pretty related/discoverable, more so than tacking it onto the nested groups example.

Not sure if you'd be interested in iterating on the above changes or if you'd rather leave that to me. I'll go ahead and start working on the above changes. Maybe it won't be too much effort :)

mwrshah:
@bvaughn I would like to take a stab at it if you don't mind. 

Agree with you, and appreciate the thought put into the advice. One thing I would push back on: freeze is sort of allowing the rest of the panels to visually alter live, while preview just moves the separator. I like how freeze looks, but it's subjective. And not entirely sure it won't add janky code to make it possible - especially that custom scrollbar hide . So I will proceed per your instructions. All good ideas: make it declarative, simllar / same separator style. not expanding docs.

bvaughn:
Happy to let you take a stab at it. Go for it!

mwrshah:
Hi! @bvaughn updated the implementation. Appreciate the time you've invested so far. I have mostly stuck to your suggestions, declaratively showing a preview during active resize, based on resizePreviewMode on Group. Reusing separator styles.

Opinionated calls taken:
- Did not introduce dedicated styling prop for the preview though it can be restyled by targeting the  data-resize-preview data attribute.
- Preview doesn't show if there is no explicit Separator, though the interaction still works by deferring resize until release. Mainly because I did not want to introduce an opinionated default style with which to show the separator

bvaughn:
Excellent! I'll take a look this morning 😄

bvaughn:
I've made a few changes to the docs (e.g. more examples) and to the API (e.g. show overlay previews for inactive separators that are impacted by a resize)

I think this feature is ready to go now! Thanks a ton for the collaboration!

PR #747, opened 2026-09-15
docs: clarify overridable inline styles in README

vadymstebakov:
## Summary

Clarifies the `style` prop documentation in the README for the `Group` and `Panel` components. It notes that default inline styles cannot be overridden, except for `overflow`, and links to an overflow example.


bvaughn (approved):

bvaughn:
Thanks

